PR #3, opened 2020-05-05
SortControl: Only apply class `icinga-controls` to the select

nilmerg:
The button doesn't require the icinga form styles.


flourish86:
I don't know, if my point became clear. The problem is shown on the screenshot.

Screenshot 2020-05-07 at 15 34 17

nilmerg:
Did you pull Icinga Web 2?

PR #4, opened 2020-05-05
Pagination control redesign

flourish86:
First shot on the implementation. Styling to be done.

Mockup here: https://github.com/Icinga/icingaweb2/issues/4111

### DIFF
--- a/src/Control/PaginationControl.php
+++ b/src/Control/PaginationControl.php
@@ -441,7 +441,7 @@ class PaginationControl extends BaseHtmlElement
         $select = Html::tag('select', [
             'name' => $this->getPageParam(),
             'class' => 'autosubmit',
-            'title' => 'Go to page …'
+            'title' => t('Go to page …')
         ]);
 
         if (isset($currentPageNumber)) {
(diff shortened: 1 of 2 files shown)


nilmerg:
Quickly glanced at it, things I noticed:

* I can't progress further than to the second and last page. (The next button has no effect after the first page)
* While I'm not on the first page I'd expect to be able to go back, but I can't.

Code review follows once the style is fixed.

src/Control/PaginationControl.php
```diff
@@ -407,8 +358,122 @@ private function translate($message)
...
+        $currentPageNumber = $this->getCurrentPageNumber();
+
+        $form = new CompatForm($this->url);
+        $form->addAttributes(['class' => 'inline']);
+        $form->setMethod('GET');
+
+        $select = Html::tag('select', [
+            'name' => $this->getPageParam(),
+            'class' => 'autosubmit',
+            'title' => 'Go to page …'
```
nilmerg, line 444: Please mark this as translatable `t(...)`

flourish86:
> Works for me now and look okay. What has now still to be styled?

Styling is contained in `controls.less` and is final so far.

PR #13, opened 2021-03-23
Implement tabs extend() method

Yonas-net:
Applies the possibility to extend tabs extensions.


nilmerg:
Will merge this now. But I'm not a fan of another legacy method being introduced in this class. Though, there's now #16 for this...

PR #25, opened 2021-06-14
Add tool tip to control items without label

raviks789:
Tool tip to LimitControl and SortControl is added to provide additional information to the user.

icinga/icingadb-web#144

### DIFF
--- a/src/Control/LimitControl.php
+++ b/src/Control/LimitControl.php
@@ -116,6 +116,7 @@ class LimitControl extends CompatForm
             'class'   => 'autosubmit',
             'label'   => '#',
             'options' => $limits,
+            'title'   => t('Change item count per page'),
             'value'   => $limit
         ]);
     }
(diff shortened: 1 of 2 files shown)


src/Control/LimitControl.php
```diff
@@ -116,6 +116,7 @@ protected function assemble()
             'class'   => 'autosubmit',
             'label'   => '#',
             'options' => $limits,
+            'title'   => sprintf('Change item count per page'),
```
flourish86, line 119: I think you should use the `t(…)` function for translation. You also don't need `sprintf()`, since you don't replace anything.

src/Control/SortControl.php
```diff
@@ -195,6 +195,7 @@ protected function assemble()
 
         $toggleButton = new ButtonElement($this->getSortParam(), [
             'class' => 'link-button spinner',
+            'title' => sprintf('Change sort direction'),
```
flourish86, line 198: Same here.

flourish86 (changes requested):

nilmerg (approved):

PR #36, opened 2021-09-06
CSS: Fix sitemap icons aren't centered

Yonas-net:
resolves https://github.com/Icinga/icingadb-web/issues/288


Yonas-net:
Bildschirmfoto 2021-09-06 um 09 44 44

flourish86 (commented):
Sorry, to say. But I don't say a difference between master and this branch on neither Firefox, Chrome or Safari.

See how the icons perfectly overlap (I overlapped screenshots from both branches and made the top one transparent):

Screen Shot 2021-09-07 at 15 29 16

Screen Shot 2021-09-07 at 15 30 00

Screen Shot 2021-09-07 at 15 30 08

Yonas-net:
I didn't get to see as many sitemap icons today, though, there' s even a big difference on my end.

**master:**

Bildschirmfoto 2021-09-07 um 17 27 02

**bugfix/fix-sitemap-icons-are-not-centered:**

Bildschirmfoto 2021-09-07 um 17 23 18

flourish86:
> I didn't get to see as many sitemap icons today, though, there' s even a big difference on my end.

Which OS?

Yonas-net:
Mac: Safari + Chrome

flourish86:
> I didn't get to see as many sitemap icons today, though, there' s even a big difference on my end.

Are you sure, this isn't a different problem, like e.g. zoom-level, then? I've just noticed, but the changes (commit d1a4e027)  address only `ball-size-l`.

The list-item's visual in default view mode contain `.ball-size-xl`.

What should be affected are the minimal list items, e.g. in the detail header, which are affected by the change (in Safari), but still misaligned.

**master**
Screen Shot 2021-09-08 at 14 19 50

**bugfix/fix-sitemap-icons-are-not-centered**
Screen Shot 2021-09-08 at 14 19 32

nilmerg:
Please also note that not all icons are affected as shown [here](https://github.com/Icinga/icingadb-web/issues/288#issuecomment-910330781). The vertical alignment is most surely just a visual thing, the icons are all still perfectly centered.

flourish86:
I found several inconsistencies, which may the origin of the problem:
* while the calculated height, line-height and font-size of the `span.icon` and `.icon:before` the same for all the different icons, the width varies dependent on the icon glyph that is used. The `sitemap` icon glyph is the widest of all. This is why it is most apparent on this one
* There seems to be a minimal font-size in Chrome and Safari, when the font-size is set in `em`s. You can have it rendered below, when you set the font-size in px. But: Browsers have the option to render a minimum font-size. Some users might have it active (See screenshot from Safari)
* As I learnt from the `align-controls` thing `display: inline-block` leads to the strongest rounding inconsistencies across browsers.

nilmerg:
@flourish86 Please take over. And don't mind to define specific rules for specific icons/glyphs that are used inside bubbles.

flourish86:
We found, that @Yonas-net changes don’t seem to be necessary.

I checked with @sukhwinder33445 on his system and we had the same results: While the icons aren’t super perfectly aligned (due to the glyphs having different shapes), I consider this good enough while being consistent. We also won’t be able to achieve this across browsers without including a browser switch, which is not something we do.

One last thing I did, was to set icons in state balls to `display: block` with 34a961e9. Again `display: inline-block` comes with several rounding inconsistencies. So we should avoid that, where possible in future.

We could have a look again, but from my point of view this can be closed.

PR #38, opened 2021-09-21
SearchBar.js: Do not open multiple modal windows on multi click

sukhwinder33445:
Related to https://github.com/Icinga/icingaweb2/pull/4538
fix https://github.com/Icinga/icingaweb2/issues/4525

### DIFF
--- a/asset/js/widget/SearchBar.js
+++ b/asset/js/widget/SearchBar.js
@@ -39,8 +39,29 @@ define(["../notjQuery"], function ($) {
             let opener = event.currentTarget;
             let editorUrl = opener.dataset.searchEditorUrl;
             let filterQueryString = this.filterInput.getQueryString();
+            let layout = document.getElementById('layout');
 
             editorUrl += (editorUrl.indexOf('?') > -1 ? '&' : '?') + filterQueryString;
+
+            // Disable pointer events to block further function calls
+            opener.style.pointerEvents = 'none';
+
+            let observer = new MutationObserver((mutations) => {
+                for (let mutation of mutations) {
+                    if (mutation.type === 'childList') {
+                        mutation.removedNodes.forEach((node) => {
+                            // Remove the pointerEvent none style to make the button clickable again
+                            // after the modal has been removed
+                            if (node.id === 'modal') {
+                                opener.style.pointerEvents = '';
+                                observer.disconnect();
+                            }
+                        });
+                    }
+                }
+            });
+
+            observer.observe(layout, {childList: true});
 
             // The search editor should open in a modal. We simulate a click on an anchor
             // appropriately prepared so that Icinga Web 2 will handle it natively.


sukhwinder33445:
When I try to fix without `setTimeout`, the pointer events are not set.

nilmerg (changes requested):
Using `setTimeout` may work, but is not reliable enough. Please use the [`MutationObserver`](https://developer.mozilla.org/de/docs/Web/API/MutationObserver) to solve this.

I would have liked to use just a new event that is triggered by the modal once closed, but that'd require users to upgrade to a new v2.9.4 or so, just to not have the opener being disabled forever. (Without a v2.9.4, tthere's no event that could invalidate the pointer events.)

Use the `MutationObserver` like this:
1) Observe `#layout` and wait for the modal being added to it
2) Observe the modal identified in 1) and wait for the css class `active` to be removed
3) Then reset the pointer events

asset/js/widget/SearchBar.js
```diff
@@ -39,9 +39,28 @@ define(["../notjQuery"], function ($) {
...
             let editorUrl = opener.dataset.searchEditorUrl;
             let filterQueryString = this.filterInput.getQueryString();
+            let layout = document.getElementById('layout');
 
             editorUrl += (editorUrl.indexOf('?') > -1 ? '&' : '?') + filterQueryString;
 
+            // Disable pointer events to block further function calls
+            setTimeout(function () {
+                opener.style.pointerEvents = 'none';
+            }, 20);
```
nilmerg, line 49: Why not instantly? Or why after 20ms?

asset/js/widget/SearchBar.js
```diff
@@ -39,9 +39,28 @@ define(["../notjQuery"], function ($) {
...
+
+            let observer = new MutationObserver(resetPointerEvents);
+            function resetPointerEvents(mutations) {
```
nilmerg, line 52: Please don't use a standard function here. It's weird that you reference it already the line beforehand and it's this way bound to the function scope. I don't know if that's reliable enough in terms of garbage collection.

Use a closure (`() => {...}`) instead.

asset/js/widget/SearchBar.js
```diff
@@ -39,9 +39,28 @@ define(["../notjQuery"], function ($) {
...
+                for (let mutation of mutations) {
+                    if (mutation.type === 'childList' && mutation.removedNodes[0]) {
```
nilmerg, line 54: You don't check if it's the modal that's being removed. Please do so.

nilmerg (changes requested):

asset/js/widget/SearchBar.js
sukhwinder33445: I have not been able to find out why, but it does not work instantly.

nilmerg: Then please try `0` for the delay. This way it's right next in the event flow, asap so to say.

sukhwinder33445: Tried without `setTimeOut` and works fine now 🤦🏽 (mysterious behaviour).

nilmerg (approved):

PR #39, opened 2021-09-22
Prevent query spam by using the searchbar

nilmerg:


nilmerg:
`LIMIT ROWS EXAMINED` could be a potential extension/alternative to this. https://mariadb.com/kb/en/limit-rows-examined/

PR #40, opened 2021-09-23
Searchbar UX improvements

### LINKED ISSUE #33: Search Editor UX Improvements
## Open the suggestions flyout, when focussing the filter input element

When the user (accidentally) closes the suggestions flyout and clicks (or focuses) the filter input again, it would be helpful to have the suggestions flyout appear again. This is how the pills behave as well.

## Consolidate `Tab` Key behavior

The tab key behaves differently depending on the suggestions flyout. 

a) With the suggestions flyout closed, `Tab` or `Shift + Tab` rotates between the pills, if at least one exists
b) When the flyout is open, `Tab` or `Shift + Tab` 

We should streamline that behavior and use the tabs for rotating between pills and have the arrow up/down keys for selecting the flyout items.

The following steps illustrate the problem

1) Go to IcingaDB > Hosts
2) Click the SearchBar to focus it
3) Type "Host" -> the flyout appears with multiple suggestions
4) Select one, e.g. "Host Name"
5) Hit enter, so the column pill for e.g. "Host Name" will be created
6) The text input will be focussed again and the flyout re-appears with suggestions for the operator. Fine so far

Behavior a)

This is the default behavior of the tab key

7) Select one of the operator suggestions -> The text input is focussed
8) Now the `Tab` (`Shift + Tab` respectively) key rotates between the pills

Behavior b)

7) The `Tab` Key now rotates through the suggestions, but there's no way to navigate back to the columns pill.

nilmerg:
resolves #33

### DIFF
--- a/asset/js/notjQuery.js
+++ b/asset/js/notjQuery.js
@@ -33,6 +33,15 @@ define(function () {
 
             if (selector === null) {
                 this.element.addEventListener(type, e => {
+                    if (type === 'focusin' && e.target.receivesCustomFocus) {
+                        // Ignore native focus event if a custom one follows
+                        if (e instanceof FocusEvent) {
+                            delete e.target.receivesCustomFocus;
+                            e.stopImmediatePropagation();
+                            return;
+                        }
+                    }
+
                     if (context === null) {
                         handler.apply(e.currentTarget, [e]);
                     } else {
--- a/asset/js/widget/BaseInput.js
+++ b/asset/js/widget/BaseInput.js
@@ -50,6 +50,7 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
             $(this.input).on('keydown', this.onKeyDown, this);
             $(this.input).on('keyup', this.onKeyUp, this);
             $(this.input).on('blur', this.onInputBlur, this);
+            $(this.input).on('focusin', this.onTermFocus, this);
             $(this.termContainer).on('input', '[data-label]', this.onInput, this);
             $(this.termContainer).on('keydown', '[data-label]', this.onKeyDown, this);
             $(this.termContainer).on('keyup', '[data-label]', this.onKeyUp, this);
--- a/asset/js/widget/BaseInput.js
+++ b/asset/js/widget/BaseInput.js
@@ -494,6 +494,13 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
             }
         }
 
+        submitTerms(terms) {
+            $(this.input.form).trigger(
+                'submit',
+                { terms: terms }
+            );
+        }
+
         moveFocusForward(from = null) {
             let toFocus;
 
@@ -775,10 +782,7 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
                 return;
             }
 
-            $(this.input.form).trigger(
-                'submit',
-                { terms: event.clipboardData.getData('text/plain') }
-            );
+            this.submitTerms(event.clipboardData.getData('text/plain'));
 
             event.preventDefault();
         }
--- a/asset/js/widget/Completer.js
+++ b/asset/js/widget/Completer.js
@@ -416,9 +416,6 @@ define(["../notjQuery"], function ($) {
                         let suggestion = suggestions[0];
 
                         this.complete(input, suggestion.value, { ...suggestion.dataset });
-                    } else if (suggestions.length) {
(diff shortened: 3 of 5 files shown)


flourish86 (approved):
Again a summary:
- [x] Open the suggestions flyout, when focussing the filter input element
- [x] Consolidate Tab Key behavior

Works nice. 👍

flourish86:
One more thing I noticed, which I struggle to understand, how it's supposed to work.

When I type something into the empty search-bar, it works as I expect:

In the hosts list:

Example 1)
a) I type some characters
b) suggestions appear
c) I don't want to select a column, but directly search for e.g. "dummy", so I select "Search for \*dummy\* in Host Name"
d) Hitting enter results in the following term: `host.name_ci=*dummy*`
e) Exactly as I'd expect and super useful ✅ 

See
Screen Recording 2021-09-29 at 16 18 24

But this doesn't work for value suggestions.

Example 2
a) I type e.g. "Host"
b) suggestions appear and I select "Host Name"
c) Enter creates a pill
d) Enter again to select the "=" operator
e) Now I type "dummy-" and suggestions appear with the first option being "Search for dummy-"
f) When I select the first option it creates a pill with "dummy-"

See
Screen Recording 2021-09-29 at 16 32 11

I think we should change f) as the first option. While it makes sense being able to enter a custom value, I suspect this isn't a helpful workflow in practice or at least not the most common.

I'd expect to have a similar behavior as in example 1 c) and have this as the first option, while the second option could be to have a custom value, as it is starting from 2. e).

Therefore 2. e) should be "Search for \*dummy\* in Host Name", while Host Name should be replaced by the column that I selected in 2. b).

nilmerg:
Do you remember our discussion about how the user should "accept" the input? Usually this is done by pushing Enter, but you suggested to make this also visible in the suggestion list. That's what the first entry is for and why it looks differently. It is accepted either by choosing it or by pushing Enter.

The quick search with the wildcards on the left and right only kicks in when defining a new condition. While defining a condition's value however, the user is expected to be already confident about what to search for and hence pushing Enter (or choosing the first suggestion) accepts exactly the chars that were typed. Making this the second option already collides with what happens when pushing Enter. And that's non-debatable because pushing Enter is *the* way in forms to confirm input.

I also really don't want to introduce multiple non-real suggestions, a single one should suffice. And this single one shows what happens when pushing Enter.

That the wildcards are not there in case of a value suggestion don't mean they're not supported, only that the user has to utilize them explicitly.

flourish86:
> Usually this is done by pushing Enter, but you suggested to make this also visible in the suggestion list. That's what the first entry is for and why it looks differently. It is accepted either by choosing it or by pushing Enter.

Yes and I'm still convinced, that this is a good idea and it's good to differentiate it visually.

> The quick search with the wildcards on the left and right only kicks in when defining a new condition. While defining a condition's value however, …

> … the user is expected to be already confident about what to search for and hence pushing Enter (or choosing the first suggestion) accepts exactly the chars that were typed. Making this the second option already collides with what happens when pushing Enter. And that's non-debatable because pushing Enter is _the_ way in forms to confirm input.

I'm sorry, but that's an assumption, that I don't see in practice.

> That the wildcards are not there in case of a value suggestion don't mean they're not supported, only that the user has to utilize them explicitly.

What speaks against displaying it as an explicit option? Especially, when we already did that for when definfing a new condition.

We highlight the default option, that is applied by pressing "Enter" by the faded `@icinga-blue` background.

From my point of view, we only gain value in having an additional option available while differentiating "non-real" values by italic font and still having the `@icinga-blue` background for the default value.

Again: Not necessarily a bug that has to be addressed immediately, but rather a feature to keep in mind for long term.

Here's a quick mockup of it (with the wording to be discussable).

Desktop

nilmerg:
> That the wildcards are not there in case of a value suggestion don't mean they're not supported, only that the user has to utilize them explicitly.

> What speaks against displaying it as an explicit option?

I didn't say that this is non-debatable. Just that I'm not a fan of this. Other opinions on the matter are welcome.

> Here's a quick mockup of it (with the wording to be discussable).

Scratch `Objects`. Scratch `in Host Name`. (that's obvious, repeated info in the same narrow area).

nilmerg:
We've concluded that the default suggestion should always be a wildcard one.

flourish86:
Works like a dream. 🤩

PR #41, opened 2021-09-23
Fix searchbar bugs

### LINKED ISSUE #32: Fix Search Editor bugs
## Clicking suggestions from the dropdown won't work (in Safari)

When clicking a suggestion from the dropdown the dropdown will be closed, but the value won't be applied. It neither works for columns, operator nor value suggestions. Selecting a value by keyboard works, though.

Screen Recording 2021-08-30 at 15 09 41

## Parsing Error when copying and pasting urls
When copying and pasting Urls with filter syntax that contain `>=` or `<=` they are parsed wrong.
* e.g. http://10.211.55.10/icingaweb2/icingadb/hosts?host.name%3E%3Dtest

**Copied url**
Screen Shot 2021-08-30 at 14 58 59

**results in**
Screen Shot 2021-08-30 at 14 59 03

Browsers: Chrome, Safari, Firefox

## Suggestion flyout exceeding viewport.
Screen Recording 2021-08-30 at 15 16 45

## Keyboard Shortcuts should work with Cmd instead of Ctrl on Mac
* Keyboard shortcuts, like `Ctrl + A` for selecting all should work with `Cmd + A`
* Copying and pasting shortcuts have no effect (at least on Mac, neither `Ctrl` nor `Cmd` will work)

## Can't deselect, after using `Ctrl + A` for selecting all
* After selecting the filter with `Ctrl + A` the filter "pills" get highlighted, but they can't be deselected.

nilmerg:
fixes #32

### DIFF
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -63,6 +63,7 @@
 .search-suggestions {
   z-index: 2; // Required so that nothing else can overlap it (such as opaque elements and the impact overlay)
   position: absolute;
+  overflow: auto;
   min-width: 5em;
 
   &:empty {
--- a/asset/js/widget/BaseInput.js
+++ b/asset/js/widget/BaseInput.js
@@ -52,7 +52,7 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
             $(this.termContainer).on('input', '[data-label]', this.onInput, this);
             $(this.termContainer).on('keydown', '[data-label]', this.onKeyDown, this);
             $(this.termContainer).on('keyup', '[data-label]', this.onKeyUp, this);
-            $(this.termContainer).on('focusout', '[data-index]', this.onTermBlur, this);
+            $(this.termContainer).on('focusout', '[data-index]', this.onTermFocusOut, this);
             $(this.termContainer).on('focusin', '[data-index]', this.onTermFocus, this);
 
             // Copy/Paste
@@ -719,7 +719,7 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
             }
         }
 
-        onTermBlur(event) {
+        onTermFocusOut(event) {
             let input = event.target;
             // skipSaveOnBlur is set if the input is about to be removed anyway.
             // If saveTerm would remove the input as well, the other removal will fail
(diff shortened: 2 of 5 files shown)


flourish86 (changes requested):
I sum up:

- [x] Clicking suggestions from the dropdown won't work (in Safari)
- [x] Parsing Error when copying and pasting urls
- [x] Suggestion flyout exceeding viewport
- [x] Can't deselect, after using Ctrl + A for selecting all


Sadly, I found an additional issue with
- [ ] Keyboard Shortcuts should work with Cmd instead of Ctrl on Mac

I can paste copied terms, when the search input is empty. But when there're terms already present, it won't parse the pasted string.

See
Screen Recording 2021-09-29 at 15 47 26
It works neither in Chrome nor Safari.

nilmerg:
> It works neither in Chrome nor Safari.

Because it's intended. Once any term is accepted/finished, the paste feature is disabled to support pasting normally.

flourish86:
> Because it's intended. Once any term is accepted/finished, the paste feature is disabled to support pasting normally.

I see the idea. Pasting is meant to be for complete filters.

I think users also may want to extend filters with existing ones or alter others they copied from somewhere else, like having the following:

a) search for a host name, e.g. "dummy-"
b) selecting the suggestion "Search for \*dummy-\* in: Host Name (CI)"
c) getting the filter-condition `host.name_ci=*dummy-*`
d) extending that by pasting one or a chain of conditions from another window or source, like `host.state.hard_state=1`

The examples are pretty simple but thinking of more complex filters, don't you think that's a practical scenario?

Some users would probably intend to do that and will be confused at the current state, when pasting behaves different for reasons that aren't comprehensible.

I know that it's rather an additional feature, than a bug. And I don't see it necessary to have all that immediately addressed for now. But in the long term, we should consider extending the search bar with that option.

nilmerg:
Hmm.. will think about it. Shouldn't be that hard to do actually.

flourish86:
> Hmm.. will think about it. Shouldn't be that hard to do actually.

Thanks, I appreciate that. 😊

PR #43, opened 2021-09-23
CSS: Remove `.search-suggestions` offset.

flourish86:
I thought, the margin was unnecessary.

### DIFF
--- a/asset/css/search-bar.less
+++ b/asset/css/search-bar.less
@@ -310,6 +310,7 @@
   }
 
   .search-suggestions {
-    margin-top: .3em;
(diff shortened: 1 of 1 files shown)


nilmerg:
It's because otherwise they overlap with the input's outline.

flourish86:
> It's because otherwise they overlap with the input's outline.

Then it should rather be [like this](https://github.com/Icinga/ipl-web/compare/bugfix/remove-suggestions-margin-top), shouldn't it?

nilmerg (approved):

PR #44, opened 2021-09-24
css: Enhance search editor style

### LINKED ISSUE #26: Fix `.search-editor` styles
- [x] The selects do have the system appearance in Safari
- [x] The select's custom arrow icon is missing
- [x] The input styles should match the `.search-bar .term`s: segmented look
- [ ] maybe match `…` buttons style with the control buttons

nilmerg:
fixes #26


asset/css/search-editor.less
```diff
@@ -18,6 +18,15 @@
     }
   }
 
+  select:not([multiple]) {
+    .appearance(none);
+    padding-right: 1.5625em;
+    background-image: url('@{iplWebAssets}/img/select-icon-white.svg');
+    background-repeat: no-repeat;
+    background-position: right center;
+    background-size: contain;
```
flourish86, line 27: `border-radius: 0` should fix the inconsistent border-radius in Safari.

Screen Shot 2021-09-29 at 14 11 14

asset/css/search-editor.less
```diff
@@ -18,6 +18,15 @@
     }
   }
 
+  select:not([multiple]) {
+    .appearance(none);
+    padding-right: 1.5625em;
+    background-image: url('@{iplWebAssets}/img/select-icon-white.svg');
```
flourish86, line 24: The icon won't load at my machine.

Screen Shot 2021-09-29 at 14 14 08

It's technically a `select`. But I'm not sure if we should actually have the icon, since we use the terms design and those don't have it.

asset/css/search-editor.less
```diff
@@ -96,7 +116,7 @@
 
     &:hover i.icon:before {
       .rounded-corners(1em);
-      background: @icinga-blue;
+      background: @gray-lighter;
```
flourish86, line 119: 👍

asset/css/search-editor.less
```diff
@@ -96,7 +116,7 @@
 
     &:hover i.icon:before {
       .rounded-corners(1em);
```
flourish86, line 118: I'd prefer
```
.rounded-corners(.25em)
```
because we don't have circular elements of this kind anywhere else in our UI.

flourish86 (changes requested):
Besides the issues, I pointed out, it looks fine.

asset/css/search-editor.less
nilmerg: restart the container for the icon to appear. The terms in the bar are not selects. But they behave like any other input in the bar and every input is auto-completed, hence why it should be okay to not have an icon there.

In the editor the input is not auto completed and doesn't behave as a text input (no cursor on focus), so the user has to know how to interact with this thing, hence the icon.

flourish86: > restart the container for the icon to appear.

Ah, got it to work. The white icon doesn't have enough contrast, though. I would rather use `select-icon.svg`, to also convey the interactivity, but this is difficult as well (remembering the "flicker contrast" thing).

See

Screen Shot 2021-09-29 at 15 11 59

So I uploaded a new one (see 65cfd8d) which seems ok.

Screen Shot 2021-09-29 at 15 13 51

> The terms in the bar are not selects. But they behave like any other input in the bar and every input is auto-completed, hence why it should be okay to not have an icon there.
> In the editor the input is not auto completed and doesn't behave as a text input (no cursor on focus), so the user has to know how to interact with this thing, hence the icon.

Agree! Sounds reasonable.

asset/css/search-editor.less
nilmerg: Except the close button in the modal right above it :laughing:

asset/css/search-editor.less
nilmerg: I was on v2.9.3 (no dark theme) at the time. With master the white icon has indeed contrast issues. Your's is way better.

asset/css/search-editor.less
nilmerg: I'll use .4em. That's the same everywhere else.

PR #52, opened 2021-11-02
Completer.js: Fix search suggestions height

sukhwinder33445:
fixes https://github.com/Icinga/icingadb-web/issues/411


asset/css/search-base.less
```diff
@@ -65,6 +65,7 @@
   position: absolute;
   overflow: auto;
   min-width: 5em;
+  margin-bottom: 2.5em;
```
nilmerg, line 68: I think this suits better in Icinga DB Web's css.

nilmerg (approved):

PR #54, opened 2021-11-04
Add `mixins.less` file

sukhwinder33445:


asset/css/mixin/mixins.less
```diff
@@ -0,0 +1,25 @@
+/*! Icinga Ipl Web | (c) 2021 Icinga GmbH | GPLv2 */
```
nilmerg, line 1: Please remove this, it's for one not required and it's not licensed under the GPL but MIT.

nilmerg (changes requested):

nilmerg (approved):
Woof, that was fast :+1:

PR #70, opened 2022-03-04
js/Add search term attributes

sukhwinder33445:


asset/js/widget/BaseInput.js
```diff
@@ -812,6 +831,20 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
+            }
+        }
+
+        isReadOnlyMode() {
+            return this.input.dataset.termMode === 'read-only'
+        }
+
+        isTermDirectionVertical() {
+            return this.input.dataset.termDirection === 'vertical'
+        }
```
nilmerg, line 846: Don't place these methods here, there are only event listeners and this section is specifically for such.

nilmerg (changes requested):
Besides my comments:

* If the term direction is vertical
  * I think pressing `Delete` or `Backspace` should have no effect *while in the main input*.
  * If `Backspace` is pressed on a term it should also be ignored, but `Delete` should probably still remove the term.
  * Pushing `Ctrl+a`, copying and pasting content should be ignored (makes no sense)
* If the terms are read only
  * The events `onKeyUp`, `onInputBlur`, `onTermFocusOut` and `onTermFocus` are completely useless and shouldn't run
* Override the methods `isReadOnlyMode` and `isTermDirectionVertical` in class `FilterInput` and return always `false` there. It isn't adjusted to support these and it makes no sense at this time anway.

asset/js/widget/BaseInput.js
```diff
@@ -666,77 +688,95 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
-                    if (input.selectionStart === input.value.length && this.hasTerms()) {
+                    if (! this.isTermDirectionVertical() && this.hasTerms()
+                        && (input.selectionStart === input.value.length || input.selectionStart === null && this.isReadOnlyMode())) {
+                        event.preventDefault();
+                        this.moveFocusForward();
+                    }
+                    break;
+                case 'ArrowUp':
+                    if (this.isTermDirectionVertical() && this.hasTerms()
+                        && (input.selectionStart === 0 || input.selectionStart === null && this.isReadOnlyMode())) {
```
nilmerg, line 766: I see no reason for checking the cursor position here. It's checked for left and right because that's how the cursor can move inside the input. This is about up and down which are directions the cursor can't move inside a single line text input anyway.

But, you should check if the input is being completed or not. Currently this and the completer compete who focuses what if there are suggestions open. Add this condition: `this.completer === null || ! this.completer.isBeingCompleted(input)`

asset/js/widget/BaseInput.js
```diff
@@ -666,77 +688,95 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
+                        event.preventDefault();
+                        this.moveFocusBackward();
+                    }
+                    break;
+                case 'ArrowDown':
+                    if (this.isTermDirectionVertical() && this.hasTerms()
+                        && (input.selectionStart === input.value.length || input.selectionStart === null && this.isReadOnlyMode())) {
```
nilmerg, line 773: Same here.

asset/js/widget/BaseInput.js
```diff
@@ -666,77 +688,95 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
-
-                            removedTerms[termIndex] = removedTerm;
-                            event.preventDefault();
                         }
-                    }
 
-                    this.togglePlaceholder();
-                    this.autoSubmit(input, 'remove', removedTerms);
+                        this.togglePlaceholder();
+                        this.autoSubmit(input, 'remove', removedTerms);
```
nilmerg, line 743: Cleanup the indentation please. You've changed only the first condition.

asset/js/widget/BaseInput.js
```diff
@@ -666,77 +688,95 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
-                        let removedTerm = this.removeTerm(input.parentNode);
-                        if (removedTerm !== false) {
-                            input = this.moveFocusForward(termIndex - 1);
-                            if (event.ctrlKey || event.metaKey) {
-                                this.clearPartialTerm(input);
-                            } else {
-                                this.writePartialTerm(input.value.slice(1), input);
+                        removedTerms = this.clearSelectedTerms();
+
+                        if (termIndex >= 0 && (! input.value || this.isTermDirectionVertical())) {
```
nilmerg, line 727: Isn't this about the read-only mode? Or why should the value be not of interest just because the direction is vertical?

asset/js/widget/BaseInput.js
```diff
@@ -666,77 +688,95 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
-
-                    if (termIndex >= 0 && ! input.value) {
-                        let removedTerm = this.removeTerm(input.parentNode);
-                        if (removedTerm !== false) {
-                            input = this.moveFocusBackward(termIndex);
-                            if (event.ctrlKey || event.metaKey) {
-                                this.clearPartialTerm(input);
-                            } else {
-                                this.writePartialTerm(input.value.slice(0, -1), input);
+                    if (! this.isTermDirectionVertical()) {
```
nilmerg, line 691: Instead of indenting the entire block further, please just break early.

asset/js/widget/BaseInput.js
nilmerg: Oh, I suggested it myself... This **is** rather about the read-only mode. I see no reason why `Delete` should be handled at all if `Backspace` isn't handled in vertical mode.

nilmerg: So, if read-only, `Backspace` should have no effect, but `Delete` no matter the value.
if the direction is vertical, both should have no effect.

asset/js/widget/BaseInput.js
```diff
@@ -49,18 +49,23 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
+                $(this.input).on('blur', this.onInputBlur, this);
+                $(this.termContainer).on('focusout', '[data-index]', this.onTermFocusOut, this);
+                $(this.termContainer).on('focusin', '[data-index]', this.onTermFocus, this);
+            }
 
             // Copy/Paste
-            $(this.input).on('paste', this.onPaste, this);
-            $(this.input).on('copy', this.onCopyAndCut, this);
-            $(this.input).on('cut', this.onCopyAndCut, this);
+            if (! this.isTermDirectionVertical()) {
```
nilmerg, line 64: I've changed my mind here. It makes no real sense for the use-case in https://github.com/Icinga/icingaweb2/pull/4691, but this is a library where this is irrelevant. Just because the direction isn't horizontal shouldn't be a reason for the copy paste functionality to not work.

asset/js/widget/BaseInput.js
```diff
@@ -49,18 +49,23 @@ define(["../notjQuery", "Completer"], function ($, Completer) {
...
             $(this.input).on('keyup', this.onKeyUp, this);
-            $(this.input).on('blur', this.onInputBlur, this);
             $(this.input).on('focusin', this.onTermFocus, this);
             $(this.termContainer).on('input', '[data-label]', this.onInput, this);
             $(this.termContainer).on('keydown', '[data-label]', this.onKeyDown, this);
-            $(this.termContainer).on('keyup', '[data-label]', this.onKeyUp, this);
-            $(this.termContainer).on('focusout', '[data-index]', this.onTermFocusOut, this);
-            $(this.termContainer).on('focusin', '[data-index]', this.onTermFocus, this);
+
+            if (! this.isReadOnlyMode()) {
```
nilmerg, line 56: I've changed my mind here as well. `onTermFocusOut` makes really no sense to run in read-only mode, that's fine.

But term selection should still be possible (especially if the copy paste functionality is active), as this is not based on whether text can be selected. So `onInputBlur` and `onKeyUp` should always run. Only `onTermFocus` needs an update, it should still deselect terms, but the syntax check and suggestion retrieval is of no use, skip that in read only mode.

nilmerg (changes requested):
Took only a look at the JS changes yet.

asset/js/widget/BaseInput.js
sukhwinder33445: Added `this.completer === null || ! this.completer.isBeingCompleted(input) && ! this.completer.hasSuggestions()`. 

! this.completer.hasSuggestions()` ensures that the suggestion list is not visible, otherwise pressing ArrowDown or ArrowUp to select the suggested options will change the focus of the terms. This will cause the selected suggestion to overwrite the existing term at position 0. 

Or if a term is being edited and something is selected from the suggestion list, a new duplicate term is added because the focus has moved again.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+use ipl\Html\HtmlElement;
+use ipl\Web\Url;
+
+/**
+ * Form for simple value based search
+ */
+abstract class SimpleSearchField extends Form
+{
+    protected $defaultAttributes = [
+        'class'                 => 'icinga-form icinga-controls search-field',
```
nilmerg, line 18: Remove `icinga-form` and `icinga-controls`. These classes are from Icinga Web 2.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+    protected $searchParameter;
+
+    /** @var Url The suggestion url */
+    protected $suggestionUrl;
+
+    /** @var string Submit label */
+    protected $submitLabel;
+
+
+
```
nilmerg, line 36: Only as single empty line please.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+use ipl\Html\Form;
+use ipl\Html\FormElement\InputElement;
+use ipl\Html\FormElement\SubmitElement;
+use ipl\Html\HtmlElement;
+use ipl\Web\Url;
+
+/**
+ * Form for simple value based search
+ */
+abstract class SimpleSearchField extends Form
```
nilmerg, line 15: Why is this abstract? It doesn't define any abstract methods. I also see no reason why it should not be used directly.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+
+
+    /**
+     * Set the search parameter to use
+     *
+     * @param   string $name
+     *
+     * @return  $this
+     */
+    public function setSearchParameter($name)
```
nilmerg, line 44: Argument and return type definitions please. Plus anywhere else here.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Get the submit label
+     *
+     * @return string
+     */
+    public function getSubmitLabel()
+    {
+        return $this->submitLabel ?? 'Submit';
```
nilmerg, line 106: That should be translated. The default.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+
```
nilmerg, line 109: Only a single empty line please

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,162 @@
...
+            'data-data-input'       => '#data-input',
+            'data-term-input'       => '#term-input',
+            'data-term-container'   => '#term-container',
+            'data-term-suggestions' => '#term-suggestions',
+            'data-suggest-url'      => $this->getSuggestionUrl()
+        ]);
+
+        $dataInput = new InputElement('data', ['type' => 'hidden', 'id' => 'data-input']);
+
+        $termInput = new InputElement('q', ['type' => 'hidden', 'id' => 'term-input']);
```
nilmerg, line 133: Shouldn't you use the set search parameter as name here?

src/Control/SimpleSuggestions.php
```diff
@@ -0,0 +1,185 @@
...
+     * @return $this
+     */
+    public function setSearchTerm($searchTerm)
+    {
+        $this->searchTerm = $searchTerm;
+
+        return $this;
+    }
+
+    /** Set the fetched data
```
nilmerg, line 43: You know what to do here :wink:

nilmerg (changes requested):
Also:

* The trash icon should only appear upon hover

src/Control/SimpleSuggestions.php
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Set the fetched data
+     *
+     * @param $data Generator
+     *
+     * @return $this
+     */
+    public function setData(Generator $data): self
```
nilmerg, line 53: `fetchSuggestions` returns `mixed`. This is incompatible. But I'm not a fan of requiring `Generator` here anyway. Remove it.

src/Control/SimpleSuggestions.php
```diff
@@ -0,0 +1,191 @@
...
+    {
+        $this->searchTerm = $searchTerm;
+
+        return $this;
+    }
+
+    /**
+     * Set the fetched data
+     *
+     * @param $data Generator
```
nilmerg, line 49: `<type> <name>` please, everywhere else as well.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,159 @@
...
+            'submit',
+            ['label' => $this->getSubmitLabel(), 'class' => 'btn-primary']
+        );
+
+        $this->registerElement($submitButton);
+
+        $this->add([
+            HtmlElement::create(
+                'div',
+                ['class' => 'control-group'],
```
nilmerg, line 153: That's Icinga Web 2.

nilmerg (changes requested):
* Pasting content auto submits, however you have disabled it, seems to be ignored
  * Also exact matches and pushing enter submits the form. This may be the default behavior, but the `TermInput` should treat this as "autosubmit" and then act depending on whether that's disabled or not.
* Stylewise, the search field should look the same as the search bar. The terms as well. The search icon should also be the same, though not necessarily have the same color. (It's not interactive after all)
  * The trash icon should apply a blur effect. The input should have a slightly darker backgorund once hovered.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        $filterInput = new InputElement(null, [
+            'type'                      => 'text',
+            'placeholder'               => 'Type to search',
+            'class'                     => 'search-input',
+            'id'                        => 'search-input',
+            'autocomplete'              => 'off',
+            'required'                  => true,
+            'data-no-auto-submit'       => true,
+            'data-no-js-placeholder'    => true,
```
nilmerg, line 120: What's this?

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,170 @@
...
+        );
+
+        $termSuggestions = new HtmlElement(
+            'div',
+            Attributes::create(['id' => 'term-suggestions', 'class' => 'search-suggestions'])
+        );
+
+        $submitButton = new SubmitElement(
+            'submit',
+            ['label' => $this->getSubmitLabel(), 'class' => 'btn-primary']
```
nilmerg, line 149: `'btn-primary'` is from Icinga Web 2 as well. You shouldn't need it anyway, you have your own style for it.

nilmerg (changes requested):
[Note to myself] Need to look at the style and js still

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,170 @@
...
+        return $this->submitLabel ?? t('Submit');
+    }
+
+    public function assemble()
+    {
+        $filterInput = new InputElement(null, [
+            'type'                      => 'text',
+            'placeholder'               => 'Type to search',
+            'class'                     => 'search-input',
+            'id'                        => 'search-input',
```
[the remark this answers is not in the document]
sukhwinder33445: I thought the form already had that class name, so I chose another, but yes.

src/Control/SimpleSearchField.php
sukhwinder33445: This disables the js placeholder. Required to display the placeholder when a term has been added and the input field is empty again.

src/Control/SimpleSearchField.php
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Form for simple value based search
+ */
+class SimpleSearchField extends Form
+{
+    protected $defaultAttributes = [
+        'class'     => 'search-field',
+        'name'      => 'search-field',
+        'role'      => 'search-field'
```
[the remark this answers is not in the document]
sukhwinder33445: Ohh, i see, it should be `search`.

src/Control/SimpleSuggestions.php
```diff
@@ -0,0 +1,190 @@
...
+        if ($request->getMethod() !== 'POST') {
+            return $this;
+        }
+
+        $requestData = json_decode($request->getBody()->read(8192), true);
+        if (empty($requestData)) {
+            return $this;
+        }
+
+        $search = str_replace(['"', "'"], '', $requestData['term']['search']);
```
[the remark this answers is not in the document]
sukhwinder33445: Yes, that was in the wrong place here, because I wanted to remove the quotes from the default suggestion. `SetDefault()` does this job now.

nilmerg:
I believe this is obsolete now. A generic input is available since #139, the vertical direction is possible since #179 and the read only mechanism will be with #222.

PR #78, opened 2022-05-05
Completer: Fix null data being emitted when triggering instrumented e…

yhabteab:
…vents

### DIFF
--- a/asset/js/widget/Completer.js
+++ b/asset/js/widget/Completer.js
@@ -240,9 +240,9 @@ define(["../notjQuery"], function ($) {
             }, 200);
         }
 
-        suggest(input, value, data = null) {
+        suggest(input, value, data = {}) {
             if (this.instrumented) {
-                if (data === null) {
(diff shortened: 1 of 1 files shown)


asset/js/widget/Completer.js
```diff
@@ -425,8 +429,9 @@ define(["../notjQuery"], function ($) {
                     break;
                 case 'Enter':
                     let defaultSuggestion = this.termSuggestions.querySelector('.default > [type="button"]');
+                    event.preventDefault();
+
                     if (defaultSuggestion !== null) {
-                        event.preventDefault();
```
nilmerg, line 429: This makes it completely impossible to submit the form even if there aren't any suggestions. Undo that, this requires a different fix in a different PR. Since I consider this an optional fix for your current project, as it also affects the search bar for a long time already and doesn't hurt, create an issue first and we discuss there/then how to fix it if there's time for it.

nilmerg (approved):

PR #79, opened 2022-05-09
Replace Filter::equal and Filter::unequal with Filter::similar and Filter::unlike correspondingly

raviks789:
depends on icinga/ipl-stdlib#33

### DIFF
--- a/src/Control/SearchEditor.php
+++ b/src/Control/SearchEditor.php
@@ -167,25 +167,45 @@ class SearchEditor extends Form
             }
 
             $newValue = $this->popKey($values, $identifier . '-value');
-            if ($newValue !== null && $rule->getValue() !== $newValue) {
+            $oldValue = $rule->getValue();
+            if ($newValue !== null && $oldValue !== $newValue) {
                 $rule->setValue($newValue);
             }
 
             $newOperator = $this->popKey($values, $identifier . '-operator');
             if ($newOperator !== null && QueryString::getRuleSymbol($rule) !== $newOperator) {
+                $value = $rule->getValue();
+                $column = $rule->getColumn();
                 switch ($newOperator) {
                     case '=':
-                        return Filter::equal($rule->getColumn(), $rule->getValue());
+                        if (is_string($value) && strpos($value, '*') !== false) {
+                            return Filter::similar($column, $value);
+                        }
+
+                        return Filter::equal($column, $value);
                     case '!=':
-                        return Filter::unequal($rule->getColumn(), $rule->getValue());
+                        if (is_string($value) && strpos($value, '*') !== false) {
+                            return Filter::unlike($column, $value);
+                        }
+
+                        return Filter::unequal($column, $value);
                     case '>':
-                        return Filter::greaterThan($rule->getColumn(), $rule->getValue());
+                        return Filter::greaterThan($column, $value);
                     case '>=':
-                        return Filter::greaterThanOrEqual($rule->getColumn(), $rule->getValue());
+                        return Filter::greaterThanOrEqual($column, $value);
                     case '<':
-                        return Filter::lessThan($rule->getColumn(), $rule->getValue());
+                        return Filter::lessThan($column, $value);
                     case '<=':
-                        return Filter::lessThanOrEqual($rule->getColumn(), $rule->getValue());
+                        return Filter::lessThanOrEqual($column, $value);
+                }
+            }
+
+            $value = $rule->getValue();
+            if ($oldValue !== $value && is_string($value) && strpos($value, '*') !== false) {
+                if (QueryString::getRuleSymbol($rule) === '=') {
+                    return Filter::similar($rule->getColumn(), $value);
+                } elseif (QueryString::getRuleSymbol($rule) === '!=') {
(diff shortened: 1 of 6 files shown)


nilmerg:
Since you're adjusting many usages of `Filter::equal` here, I think most of them are not necessary.

Consider the query `foo=bar`. This is an equal comparison. There's no wildcard, so why should it be handled as a `Similar` condition? ipl-sql already renders `Similar` conditions the same as `Equal` conditions if there's no wildcard.

The filter renderer/parser should handle this the same I think. If there's a wildcard, it can only be a `Similar` or `Unlike` condition. If there's no wildcard, it's an `Equal` or `Unequal` condition.

This way the search bar and other components need to support all types. And can still use `Equal` in places where `Similar` is not required.

And don't forget the tests here :wink:

src/Control/SearchEditor.php
```diff
@@ -173,19 +173,38 @@ protected function applyChanges(Filter\Rule $rule, array &$values, array $path =
...
+            }
+
+            $value = $rule->getValue();
+            if (is_string($value) && strpos($value, '*') !== false) {
+                if (QueryString::getRuleSymbol($rule) === '=') {
+                    return Filter::similar($rule->getColumn(), $value);
+                } elseif (QueryString::getRuleSymbol($rule) === '!=') {
+                    return Filter::unlike($rule->getColumn(), $value);
                 }
             }
```
nilmerg, line 209: At first I didn't knew what this section is for. After thinking about it, I do: Because the value can also change from non-wildcard to wildcard. Though, it would make more sense if you'd place this section inside the structure at line no. 170. It makes the intention of the code more clear, and it doesn't run needlessly. Though, you can't just return a new rule there, the operator might still change after all. So maybe don't move it, but also check here if the value has changed and don't replace the rule needlessly.

tests/FilterTest.php
```diff
@@ -23,6 +23,25 @@ public function testParserIdentifiesSingleConditions()
         );
     }
 
+    public function testParserIdentifiesWildCardCharacters()
```
nilmerg, line 26: This test is identical to the new cases in `testParserIdentifiesRelationalOperators`. Why not change it to check if the parser correctly differentiates between the condition types if there's a wildcard and if there's none?

tests/FilterTest.php
```diff
@@ -340,6 +373,9 @@ public function testParserIdentifiesTheHolyGrail()
             /* testParserIdentifiesBooleanConditions */
             Filter::equal('active', true),
             Filter::equal('active', false),
+            /* testParserIdentifiesWildCardCharacters */
```
nilmerg, line 376: These two are rather a part of `testParserIdentifiesRelationalOperators`

nilmerg (changes requested):

src/Control/SearchEditor.php
raviks789: Did not move this block but added a condition, so that the rule changes only when the value has been changed.

tests/FilterTest.php
raviks789: Changed the test as per your suggestion.

tests/FilterTest.php
```diff
@@ -23,6 +23,37 @@ public function testParserIdentifiesSingleConditions()
         );
     }
 
+    public function testParserIdentifiesWildCardCharacters()
+    {
+        $expectedSimilar = QueryString::parse("foo=ba*");
+
+        $this->assertTrue(
```
nilmerg, line 30: I'm sure there are special type assertions in php-unit ;)

src/Control/SearchEditor.php
```diff
@@ -167,25 +167,45 @@ protected function applyChanges(Filter\Rule $rule, array &$values, array $path =
...
 
             $newValue = $this->popKey($values, $identifier . '-value');
+            $oldValue = $rule->getValue();
             if ($newValue !== null && $rule->getValue() !== $newValue) {
```
nilmerg, line 171: Maybe access `$oldValue` here as well. :D

src/Control/SearchEditor.php
```diff
@@ -167,25 +167,45 @@ protected function applyChanges(Filter\Rule $rule, array &$values, array $path =
...
-                        return Filter::lessThan($rule->getColumn(), $rule->getValue());
+                        return Filter::lessThan($column, $value);
                     case '<=':
-                        return Filter::lessThanOrEqual($rule->getColumn(), $rule->getValue());
+                        return Filter::lessThanOrEqual($column, $value);
+                }
+            }
+
+            $value = $rule->getValue();
+            if (is_string($value) && strpos($value, '*') !== false && $oldValue !== $value) {
```
nilmerg, line 204: Check `$oldValue !== $value` first, the other checks don't run then if that's not true.

nilmerg (changes requested):
Just some minor nitpicks, thanks so far! :)

tests/FilterTest.php
raviks789: Used `assertInstanceOf()`, I guess that's what you wanted me to use.

nilmerg (approved):

PR #101, opened 2022-10-25
Introduce `ScheduleElement`

yhabteab:


yhabteab:
### Blocked by

- https://github.com/Icinga/ipl-html/pull/74
- https://github.com/Icinga/ipl-html/pull/68

flourish86:
* Please preselect the option “Each” here and regardlessly disable the inputs for the one, that is not selected.
Screenshot 2022-11-17 at 16 46 24

* Also for annually … 
Screenshot 2022-11-17 at 16 47 24

* Besides that, either remove or enable the “Every field …”.

yhabteab:
> I’m not sure, if this is a bug, but the occurences are displayed wrong or not set correctly:

You need to checkout this branch as well. https://github.com/Icinga/ipl-scheduler/pull/12

> Also the seconds seem to be ignored.

That's to be expected!

asset/css/crontab.less
```diff
@@ -0,0 +1,129 @@
...
+form.icinga-form .control-group.crontab-wrapper, .monthly-annually-selector {
+  margin-bottom: 0;
+}
+
+form.icinga-form .control-group.crontab-wrapper {
+  margin-top: -1em;
+}
+
+form.icinga-form .crontab {
+  flex: 2;
```
nilmerg, line 21: Why not just `.crontab`?
Where does this become a flex item? Layout CSS should be self-contained and dependent rules should be applied near the same location and if possible even in a combined rule block.

bad:

```css
.foo {
  display: flex;
}

.bar { // Happens to be inside .foo
  flex: 1 1 auto;
}
```

better:

```css
.foo {
  display: flex;
}

.foo > .bar {
  flex: 1 1 auto;
}
```

perfect: (LESS)

```css
.foo {
  display: flex;

  > .bar {
    flex: 1 1 auto;
  }
}
```

nilmerg (changes requested):
Additionally:

* Please use the LESS variables of this library, not these by Icinga Web
* Avoid unnecessary specificity
  * If unavoidable, keep the ruleset compact and put other rules (e.g. style) in less specific rulesets

yhabteab:
I don’t know why but when using `@base-disabled` instead of `@disabled-gray` as a text color doesn’t seem to work.
Bildschirm­foto 2022-11-18 um 16 22 54

nilmerg:
I don't know as well, but if you take a look at the notice at the start of `variables.less` you can see you shouldn't use this directly anyway.

flourish86:
I’m not sure about the markup here, since it’s not valid html having different items than `li` in a `ul`.

Screenshot 2022-11-21 at 10 42 31

I think it’s totally possible to do it like this.

```
<ul>
  <li>
    <label>
      <input type=“checkbox”>
      …
    <label>
  </li>
…
</ul>
```

yhabteab:
refs https://github.com/Icinga/icingaweb2/pull/4946

bobapple:
@cla-bot check

nilmerg (commented):
Markup looks good so far.

asset/css/crontab.less
```diff
@@ -0,0 +1,107 @@
...
+    padding: .4em;
+    border: .1em solid @crontab-list-border-color;
+    .rounded-corners(.6em);
+
+    li {
+      margin-bottom: .08em;
+    }
+  }
+
+  .list-items {
```
nilmerg, line 43: I believe that `list-items` is a way too common class name. What about e.g. `.monthly > ul`?

asset/css/crontab.less
```diff
@@ -0,0 +1,107 @@
...
+        cursor: pointer;
+        text-align: center;
+        padding: .8em 0 .8em;
+        border-right: .1em solid @crontab-list-border-color;
+        background: @crontab-selectable-list-bg;
+        color: @crontab-default-text-color;
+      }
+
+      &:hover {
+        background-color: @crontab-list-hover-color;
```
nilmerg, line 72: This has no effect for me, the color is near transparent. Also, please think of keyboard only users which use TAB to navigate. Add `:focus` here to also show them what they've selected.

asset/css/variables.less
```diff
@@ -100,6 +100,13 @@
 
 @card-border-color: @base-gray-light;
 
+@crontab-disabled-text-color: @base-disabled;
+@crontab-selectable-list-bg: #404d72;
```
nilmerg, line 104: `@searchbar-bg` uses the same color code. I think it's time now for a `@default-input-bg`.

nilmerg (changes requested):
Besides my comments, the CSS looks better already. Nice job!

asset/css/crontab.less
yhabteab: These rules apply not only to monthly lists, but also to weekly and yearly. However, I could prefix the class name with `crontab`, but I don't know which version is preferable!

src/Widget/Crontab.php
```diff
@@ -0,0 +1,634 @@
...
+    }
+
+    /**
+     * Set callback to protect ids with
+     *
+     * @param callable $protector
+     *
+     * @return $this
+     */
+    public function setProtector($protector): self
```
nilmerg, line 122: Please name this `setIdProtector`. That's the name in use any where else.

nilmerg (changes requested):
I'm not sure if this is really a *widget*. It's useless without a form and extends `FieldsetElement`. So, in essence this is a form element. Albeit a rather complex one. Though, that's why it's here in ipl-web, not in ipl-html. But that doesn't mean it can't be a form element. Please introduce a new namespace `ipl\Web\FormElement` and place this there.

While you're relocating the file, think about a name change as well. `Crontab` is too technical for me. *Cron* is only an implementation detail. What about `ScheduleElement`? The method `load` should then indicate that it loads a cron expression and be named `loadCron`. `getPreparedExpression` should indicate this as well and be named `getCronExpression`. Don't forget to adjust the CSS classes. :wink:

nilmerg (commented):
css looks good again

flourish86 (commented):
As discussed (offline) we should change the following:

1) The annual and monthly scheduler elements work differently (multiselect vs. single select), so this should be made clear by the design.

2) Also the “On the …” option for annual scheduler should be a checkbox type input. In opposition to the monthly scheduler it is an addition to the month selection and not a different option.

3) We’ll have to find a different solution for the apply button, probably with autosubmit.

4) The “Day” option in “On the” only makes sense for day, since all the other options would be ambiguous und thus not renderable.

5) Let’s change the option labels for Weekday to “Weekday (Mon – Fri)” and “Weekend (Sat + Sun)” so it’s clear, what it does.

I will look for design solutions for 1) and 2).

flourish86:
> As discussed (offline) we should change the following:
>
> 2. Also the “On the …” option for annual scheduler should be a checkbox type input. In opposition to the monthly scheduler it is an addition to the month selection and not a different option.

Desktop Copy 18

flourish86 (commented):
1) When there’s an item selected in the annual scheduler and I switch to “monthly”, I get an error.
```
Uncaught Error: Call to undefined method ipl\Web\FormElement\ScheduleElement::hasBeenSubmitted() in /usr/local/src/ipl-html/src/FormElement/FormElements.php:354
```

2) The monthly scheduler is out of place now.
Screenshot 2022-12-08 at 16 43 42

3) “Next occurences” arent correct.
Screenshot 2022-12-08 at 16 44 26

The rest looks fine so far.

asset/css/schedule-element.less
```diff
@@ -0,0 +1,164 @@
...
+      margin-right: 1px;
+      margin-bottom: 1px;
+    }
+
+    .schedule-element-fields li label {
+      border-right: none;
+    }
+
+    .schedule-element-fields:focus-within {
+      outline: 3px solid rgba(0, 195, 237, 0.5);
```
nilmerg, line 148: Can we outsource the rgba value to a variable?

nilmerg (changes requested):

nilmerg:
Is this only for me with Chromium or is the focus navigation of single select fields generally impossible?

asset/css/schedule-element.less
flourish86: I used a function to fade the `@base-primary-bg`.

flourish86:
> Is this only for me with Chromium or is the focus navigation of single select fields generally impossible?

It might be just you.

see (Brave, though). I’m tabbing with Tab, arrow keys won’t work.


https://user-images.githubusercontent.com/6977434/212920258-0a82d56d-d978-4e50-a731-1b0f6bbda1bc.mov

yhabteab:
> > Is this only for me with Chromium or is the focus navigation of single select fields generally impossible?
> 
> It might be just you.

The single select is only available for the `annually` schedules, and it's intended that you can’t focus on the individual fields, but rather the parent wrapper.

nilmerg:
And how do I choose a month with the keyboard?

yhabteab:
You can do this by using the arrows and not tabs

nilmerg:
Aaaah, because these are radio elements! :bulb:

nilmerg:
Though, what this shows, I wonder whether this is intuitive. They don't look like radio options after all. :thinking:

flourish86:
> Though, what this shows, I wonder whether this is intuitive. They don't look like radio options after all. 🤔

We’ve regarded this with the difference in appearance. The multiselect inputs also don’t look like checkboxes. Though, I have to admit, that you can argue that they are more broadly known, of course.

The difference in keyboard interactions makes these pretty unintuitive. But that’s how the keyboard interaction works by default, sadly. I guess the different behaviors for checkboxes and radio buttons isn’t pretty obvious.

So, what would the options to solve this:

1) Display checkboxes and radios?
You win your (maybe) known behavior as a keyboard user. But all users lose clarity, since the additional repeating elements distract from the important content. Every altered version of the original appearance would probably not be helpful.

2) Create a custom behavior to make them behave more intuitive with keyboard interaction?
This would be the most user friendly option. Ideally you should be able to use arrow key interaction in all four directions, which we could also address with that.

3) Display a hint, when keyboad interaction is initiated?
Helpful for keyboard users, but another option requiring a custom behavior.

For me the first option would be out, since it would do more harm than it’ll be helpful.

What do you think?

asset/css/variables.less
```diff
@@ -162,5 +179,14 @@
     --suggestions-relation-path-focus-bg: var(--base-gray);
 
     --card-border-color: var(--base-gray-light);
+
+    --schedule-element-fields-bg: var(--default-input-bg);
+    --schedule-element-default-text-color: #F5F9FA;
+
+    --schedule-element-fields-hover-bg: #C7E8F0;
+    --schedule-element-fields-selected-hover-bg: #6FD7EF;
```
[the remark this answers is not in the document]
flourish86: ✅

nilmerg:
Option 3. But without JS. Isn't it an option to show an indicator in the top right corner (overlayed over the element there, showing an icon and/or a short "Use <icon> to move") for example upon focus?

I have only one type of user in mind with this. The one preferring the keyboard over the mouse. Users with screenreaders don't need this as the screenreader knows it's a set of radio buttons.

flourish86:
> Option 3. But without JS. Isn't it an option to show an indicator in the top right corner (overlayed over the element there, showing an icon and/or a short "Use <icon> to move") for example upon focus?

I like the idea, but I wouldn’t overlay anthing. I could image though, that it looks something like this.

Positioning it below the schedule-element is what matches the use case best. The user won’t be interested before actually being able to use the element. So, having it close to the actual element would be the best solution.

We could have it there by default or (if possible) have it "pop up”, i.e. set to display, when the schedule element is focussed.

Artboard Copy

nilmerg:
Oof, a bit too prominent for my taste. And upon focus only of course, but without letting the other elements *bounce*, that's why I'd opt for an overlay. `Use ⇆ to navigate` or `Choose by ⇆` should be enough. A combined icon exists. It should be obvious that the keyboard is used for this.

src/FormElement/ScheduleElement.php
```diff
@@ -499,6 +502,16 @@ private function assembleMonthlyOrAnnuallySchedules(): void
...
 
         $ordinalWrapper->addHtml($enumerations, $selectableDays);
+
+        if ($isAnnually) {
+            $fieldsWrapper->add(HtmlElement::create('div', ['class' => 'note'], [
+                'Use ',
+                new Icon('arrow-left'),
+                ' / ',
+                new Icon('arrow-right'),
+                ' keys to navigate the active element with keyboard.'
```
nilmerg, line 512: 1. This is not translatable
2. ~~navigate the active element with~~ *choose a month by*

asset/css/variables.less
```diff
@@ -100,6 +103,21 @@
...
 @card-border-color: @base-gray-light;
 
+@schedule-element-fields-bg: @default-input-bg;
+@schedule-element-default-text-color: @default-bg;
+@schedule-element-text-color: @base-primary-color;
+
+@schedule-element-fields-selected-bg: @base-primary-color;
+@schedule-element-fields-hover-bg: fade(@base-primary-dark, 30);
+
+@schedule-element-fields-selected-hover-bg: #00A5CF;
```
nilmerg, line 113: This is still initialized with an absolute value.

asset/css/variables.less
```diff
@@ -100,6 +103,21 @@
 
 @card-border-color: @base-gray-light;
 
+@schedule-element-fields-bg: @default-input-bg;
+@schedule-element-default-text-color: @default-bg;
```
nilmerg, line 107: text-color != bg

asset/css/variables.less
```diff
@@ -100,6 +103,21 @@
...
+@schedule-element-text-color: @base-primary-color;
+
+@schedule-element-fields-selected-bg: @base-primary-color;
```
nilmerg, line 110: bg != text-color

asset/css/variables.less
```diff
@@ -162,5 +181,13 @@
     --suggestions-relation-path-focus-bg: var(--base-gray);
 
     --card-border-color: var(--base-gray-light);
+
+    --schedule-element-fields-bg: var(--default-input-bg);
+    --schedule-element-default-text-color: var(--base-primary-color);
```
nilmerg, line 186: Interesting. In the dark theme it's initalized with `@default-bg`. Why here not with `var(--default-bg)` as well? I mean, please not with `default-bg` again, but with the same variable counterpart please!

asset/css/variables.less
```diff
@@ -162,5 +181,13 @@
...
+
+    --schedule-element-fields-hover-bg: @base-primary-light;
```
nilmerg, line 188: In light mode, using a less variable is the same as an absolute value in dark mode. Please avoid this on specific variables.

nilmerg (changes requested):

asset/css/variables.less
flourish86: 💡✅

asset/css/variables.less
flourish86: I don’t wanna go uneccessarily deep into this, but I’d rather opt for `@default-body-bg` or `@default-viewport-bg` to be less ambiguous for both.

asset/css/variables.less
flourish86: (v)

asset/css/variables.less
flourish86: ✅

asset/css/variables.less
flourish86: ✅

src/FormElement/ScheduleElement.php
flourish86: Seems already adressed by 6d973be.

flourish86 (dismissed):
Couldn’t find any ui related issues.

asset/css/variables.less
```diff
@@ -162,5 +182,13 @@
...
 
     --card-border-color: var(--base-gray-light);
+
+    --schedule-element-fields-bg: var(--default-input-bg);
+    --schedule-element-default-text-color: var(--base-primary-color);
+
+    --schedule-element-fields-hover-bg: var(--base-primary-dark);
+
+    --schedule-element-fields-border-color: var(--base-gray-light);
+    --schedule-element-fields-disabled-bg: var(--base-gray-lighter);
```
nilmerg, line 192: I still miss the following definitions:

* schedule-element-text-color
* schedule-element-fields-selected-bg
* schedule-element-fields-selected-hover-bg
* schedule-element-fields-disabled-color
* schedule-element-fields-disabled-selected-bg

Remember, if a LESS variable is defined with a base/default variable in dark mode, the same must be done in light mode.

nilmerg (changes requested):
@yhabteab The radio options for the monthly fields do not disable the other group's elements anymore.

yhabteab:
> @yhabteab The radio options for the monthly fields do not disable the other group's elements anymore.

Ohh, yeah. The `preg_match()` pattern for the multipart updates is broken.

yhabteab:
> > @yhabteab The radio options for the monthly fields do not disable the other group's elements anymore.
> 
> Ohh, yeah. The `preg_match()` pattern for the multipart updates is broken.

Is fixed here https://github.com/Icinga/ipl-web/pull/108

asset/css/variables.less
```diff
@@ -48,7 +50,7 @@
 
 @primary-button-color: @default-text-color-inverted;
 @primary-button-bg: @base-primary-bg;
-@primary-button-hover-bg: #0081a6;
+@primary-button-hover-bg: #00A5CF;
```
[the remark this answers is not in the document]
flourish86: ✅

asset/css/variables.less
```diff
@@ -109,15 +126,18 @@
 
     --base-remove-bg: @state-critical;
 
+    --base-primary-dark: fade(@primary-button-bg, 35%);
```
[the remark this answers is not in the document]
flourish86: ✅

asset/css/variables.less
flourish86: ✅

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -0,0 +1,141 @@
...
+     * Set the default weekday to be preselected
+     *
+     * @param string $default
+     *
+     * @return $this
+     */
+    public function setDefault(string $default): self
+    {
+        $weekday = strlen($default) > 2 ? substr($default, 0, -1) : $default;
+        // Attributes are registered far before the initialization of this element!
```
nilmerg, line 46: This wouldn't be necessary if you initialize `$this->weekdays` in the constructor, right? Since this class is internal (That's a PHPdoc annotation by the way :wink: ), overriding the constructor does no harm.

asset/css/variables.less
```diff
@@ -162,5 +182,13 @@
     --suggestions-relation-path-focus-bg: var(--base-gray);
 
     --card-border-color: var(--base-gray-light);
+
+    --schedule-element-fields-bg: var(--default-input-bg);
+    --schedule-element-default-text-color: var(--base-primary-color);
```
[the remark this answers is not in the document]
flourish86: ✅

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $this->getAttributes()->set('id', $this->protectId('weekly-fields'));
+
+        $fieldsWrapper = HtmlElement::create('div', ['class' => 'weekly']);
+        $listItems = HtmlElement::create('ul', ['class' => ['schedule-element-fields', 'multiple-fields']]);
+
+        $foundCheckedDay = false;
+        foreach ($this->weekdays as $day => $value) {
+            $checkbox = $this->createElement('checkbox', $day, [
+                'class' => 'sr-only autosubmit',
```
nilmerg, line 109: If you define multiple attribute values this way, please use an array. Any subsequent call to `$attribute->add()` then behaves correctly. This is not that important here now, but it's a good habit to remember this and not use a single string for multiple values.

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -0,0 +1,141 @@
...
+                'value' => $this->getPopulatedValue($day, 'n')
+            ]);
+            $this->registerElement($checkbox);
+
+            $foundCheckedDay = $foundCheckedDay || $checkbox->isChecked();
+            $htmlId = $this->protectId("weekday-$day");
+            $checkbox->getAttributes()->set('id', $htmlId);
+
+            $listItem = HtmlElement::create('li');
+            $checkbox->prependWrapper($listItem);
```
nilmerg, line 119: You don't use `addElement` above, why is this necessary? Just add `$listItem` to `$listItems` below. I find it irritating anyway that the checkbox is part of `$listItem` **and** `$listItems`.

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -0,0 +1,141 @@
...
+
+            $listItem->addHtml($checkbox, HtmlElement::create('label', ['for' => $htmlId], $value));
+            $listItems->addHtml($checkbox);
+        }
+
+        if (! $foundCheckedDay) {
+            $this->getElement($this->default)->setChecked(true);
+        }
+
+        $listItems->prependWrapper($fieldsWrapper);
```
nilmerg, line 129: You should avoid using a wrapper here. Just add `$listItems` to `$fieldsWrapper` and add the latter to `$this`.

nilmerg (changes requested):

src/FormElement/ScheduleElement/AnnuallyFields.php
```diff
@@ -0,0 +1,145 @@
...
+        $annuallyWrapper->addHtml($fieldsSelector);
+
+        $notes = HtmlElement::create('div', ['class' => 'note']);
+        $notes->addHtml(
+            HtmlString::create($this->translate('Use ')),
+            new Icon('arrow-left'),
+            HtmlString::create(' / '),
+            new Icon('arrow-right'),
+            HtmlString::create($this->translate(' keys to choose a month by keyboard.'))
+        );
```
nilmerg, line 112: It's now translatable, but: Imagine how a translator sees this:

* `Use `
* ` keys to choose a month by keyboard.`

Take a look [here](https://translate.icinga.com/translate/icinga/icingadb-web/de/?q=) and imagine these intermingled with others in between. Yeah, they are grouped by location and appearance in files, but we shouldn't make it harder than required.

Construct messages like this with sprintf syntax:

`FormattedString::create($this->translate('Use %s / %s keys to choose a month by keyboard.'), <icon>, <icon>)`

nilmerg (changes requested):

src/FormElement/ScheduleElement/AnnuallyFields.php
flourish86: (v)

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,81 @@
...
+    protected $disable;
+
+    /**
+     * En/Disable the "on the" radio options of this element
+     *
+     * @param bool $value
+     *
+     * @return $this
+     */
+    public function disable(bool $value): self
```
nilmerg, line 25: Is this unused?

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,81 @@
...
+        foreach ($this->options as $option) {
+            $radio = (new InputElement($this->getValueOfNameAttribute()))
+                ->setValue($option->getValue())
+                ->setType($this->type);
+
+            $radio->setAttributes(clone $this->getAttributes());
+
+            $htmlId = $this->protectId($option->getValue());
+            $radio->getAttributes()
+                ->registerAttributeCallback('id', function () use ($htmlId) {
```
nilmerg, line 44: Why is this a callback?

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,81 @@
...
+                    return $htmlId;
+                })
+                ->registerAttributeCallback('checked', function () use ($option) {
+                    return (string) $this->getValue() === (string) $option->getValue();
+                })
+                ->registerAttributeCallback('required', function () {
+                    return $this->getRequiredAttribute();
+                })
```
nilmerg, line 52: Use `[$this, 'getRequiredAttribute']`

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,81 @@
...
+                ->registerAttributeCallback('disabled', function () use ($option) {
+                    return $option->isDisabled();
+                })
```
nilmerg, line 55: `[$this, 'isDisabled']`, I guess

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,81 @@
...
+                ->registerAttributeCallback('class', function () use ($option) {
+                    return Attributes::create(['class', $option->getLabelCssClass()])->get('class');
```
nilmerg, line 57: Doesn't just `$option->get...` suffice?

nilmerg (changes requested):

nilmerg (changes requested):
Switching between the months in annually mode flickers for me in Chromium pretty hard. I believe this is because the focus is lost first and then shortly after gained by another element. This transition may be the reason for the flickering. Can we avoid this with a transition (with a delay) at least on the note?


https://user-images.githubusercontent.com/16668527/216374019-04c9bf87-f796-444e-87c6-31ed7bc0c962.mp4

flourish86:
> Switching between the months in annually mode flickers for me in Chromium pretty hard. I believe this is because the focus is lost first and then shortly after gained by another element. This transition may be the reason for the flickering. Can we avoid this with a transition (with a delay) at least on the note?
> 
>  Screencast.2023-02-02.16.52.54.mp4

I can’t reproduce that on Mac.

It looks more of a problem with a delay in the focus being changed. I don’t think a transition could fix that.

https://user-images.githubusercontent.com/6977434/216379766-2a493705-0e8e-48c9-a76e-6b81216de79e.mov

src/FormElement/ScheduleElement/AnnuallyFields.php
```diff
@@ -0,0 +1,149 @@
...
+
+        return $this;
+    }
+
+    protected function assemble()
+    {
+        $this->getAttributes()->set('id', $this->protectId('annually-fields'));
+
+        $fieldsSelector = new FieldsRadio('month', [
+            'class'     => 'autosubmit sr-only',
```
nilmerg, line 85: :wink:

src/FormElement/ScheduleElement/AnnuallyFields.php
```diff
@@ -0,0 +1,149 @@
...
+
+        $runsOnThe = $this->getPopulatedValue('runsOnThe', 'n');
+        $this->addElement('checkbox', 'runsOnThe', [
+            'class' => 'autosubmit',
+            'value' => $runsOnThe
+        ]);
+
+        $checkboxControls = HtmlElement::create('div', ['class' => 'toggle-slider-controls']);
+        $checkbox = $this->getElement('runsOnThe');
+        $checkbox->prependWrapper($checkboxControls);
```
nilmerg, line 102: I'm confident that you can avoid using wrappers here as well :wink:

nilmerg (changes requested):

src/FormElement/ScheduleElement/AnnuallyFields.php
yhabteab: How? The checkbox is added using `addElement()`.

yhabteab:
There is no focus for the annually fields on safari and navigating using the arrows doesn't work either.
Bildschirm­foto 2023-02-03 um 10 32 57

src/FormElement/ScheduleElement/AnnuallyFields.php
nilmerg: Well, I missed that of course -.- now I'm not that confident anymore :rofl:

nilmerg (changes requested):

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -0,0 +1,175 @@
...
+            'class'    => 'autosubmit',
+            'options'  => [static::RUNS_ONTHE => $this->translate('On the')]
+        ]);
+
+        $runsOnThe = $this->getElement('runsOn');
+        $runsOnValidators = $runsOnThe->getValidators();
+        $runsOnValidators
+            ->clearValidators()
+            ->add(
+                new DeferredInArrayValidator(function (): array {
```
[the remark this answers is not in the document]
yhabteab: Because I want to postpone the validation for later.

Bildschirm­foto 2023-02-03 um 14 32 21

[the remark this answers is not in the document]
yhabteab: Never mind! I have to set the `haystack` key explicitly.

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -0,0 +1,175 @@
...
+        $runsEach->prependWrapper($monthlyWrapper);
+        $monthlyWrapper->addHtml($runsEach, $listItems);
+
+        $this->addElement('radio', 'runsOn', [
+            'required' => $runsOn !== static::RUNS_EACH,
+            'class'    => 'autosubmit',
+            'options'  => [static::RUNS_ONTHE => $this->translate('On the')]
+        ]);
+
+        $runsOnThe = $this->getElement('runsOn');
```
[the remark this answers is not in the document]
yhabteab: I have to! These are not checkboxes but radio elements, and I need to render two radio options with the same name in different places.

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -0,0 +1,175 @@
...
+        if (! $foundCheckedDay) {
+            $this->getElement("day{$this->default}")->setChecked(true);
+        }
+
+        $monthlyWrapper = HtmlElement::create('div', ['class' => 'monthly']);
+        $runsEach = $this->getElement('runsOn');
+        $runsEach->prependWrapper($monthlyWrapper);
+        $monthlyWrapper->addHtml($runsEach, $listItems);
+
+        $this->addElement('radio', 'runsOn', [
```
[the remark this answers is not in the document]
yhabteab: Yeah, but is rendered in different place.

src/FormElement/ScheduleElement/Recurrence.php
```diff
@@ -0,0 +1,96 @@
...
+        if (! $isValid) {
+            return;
+        }
+
+        /** @var RRule $frequency */
+        $frequency = ($this->frequencyCallback)();
+        $recurrences = $frequency->getNextRecurrences(new DateTime(), 3);
+        if (! $recurrences->valid()) {
+            // Such a situation can be caused by setting an invalid end time
+            $this->addHtml(HtmlString::create($this->translate('Recurrences cannot be generated')));
```
nilmerg, line 68: That's not a HTML string. It's text. Use `ipl\Html\Text`. Otherwise, a malicious translation is not escaped.

src/FormElement/ScheduleElement/Recurrence.php
```diff
@@ -0,0 +1,96 @@
...
+        foreach ($recurrences as $recurrence) {
+            $this->addHtml(
+                HtmlElement::create(
+                    'p',
+                    null,
+                    sprintf(
+                        '%s, %s',
+                        $recurrence->format('D'),
+                        $recurrence->format('Y/m/d, H:i:s')
+                    )
```
nilmerg, line 82:
```suggestion
                    $recurrence->format($this->translate('D, Y/m/d, H:i:s'))
```

nilmerg (changes requested):

src/FormElement/ScheduleElement/MonthlyFields.php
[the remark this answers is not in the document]
nilmerg: No you don't. Register one option, override its validator to accept both choices. Only addHtml the other one.

src/FormElement/ScheduleElement/Common/FieldsUtils.php
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    protected function createOrdinalSelectableDays(): FormElement
+    {
+        return $this->createElement('select', 'day', [
+            'class'   => 'autosubmit',
+            'value'   => $this->getPopulatedValue('day', static::$everyDay),
+            'options' => [
+                'Regular'       => $this->regulars,
+                'Non Standards' => [
```
nilmerg, line 66: Please translate the keys.

nilmerg: Or, why are they of advantage? Is it important for the user what regular/non-standard means? Or can the subsections just be removed?

src/FormElement/ScheduleElement/Common/FieldsUtils.php
```diff
@@ -0,0 +1,241 @@
...
+                    $months[] = $rule->getStartDate()->format('m');
+                } elseif (empty($months)) {
+                    $months[] = date('m');
+                }
+
+                $values['month'] = strtoupper($this->getMonthByNumber((int)$months[0]));
+                $values['runsOnThe'] = ! empty($byDay) ? 'y' : 'n';
+            }
+
+            if (count($byDay) == 1 && preg_match('/^(-?\d)+(\S.*)$/', $byDay[0], $matches)) {
```
nilmerg, line 110: `(-?\d+)`?

It's slightly more efficient.

nilmerg (changes requested):

src/FormElement/ScheduleElement/Common/FieldsUtils.php
yhabteab: I just want to separate them. We could also do this by just rendering a line but I don't know how.

src/FormElement/ScheduleElement/Common/FieldsUtils.php
yhabteab: But it doesn't match what I would like to do. Test string: `-1MO` or `2SA`

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -0,0 +1,173 @@
...
+            'required' => $runsOn !== static::RUNS_EACH,
+            'class'    => 'autosubmit',
+            'options'  => [static::RUNS_ONTHE => $this->translate('On the')]
```
nilmerg, line 135: This also works. Saves you the get/clear/add shenanigans below. Yes, break chain on failure can't be defined, but it's a single validator anyway ;)

```suggestion
            'options'  => [static::RUNS_ONTHE => $this->translate('On the')],
            'validators' => [new InArrayValidator([
                'strict'   => true,
                'haystack' => [static::RUNS_EACH, static::RUNS_ONTHE]
            ])]
```

src/FormElement/ScheduleElement/Common/FieldsUtils.php
nilmerg: It replaces only `(-?\d)+` of course :wink:

src/FormElement/ScheduleElement/Common/FieldsUtils.php
nilmerg: Labels should have a purpose. If there's none or little, they should be replaced/removed. You could add an option with some dashes or a unicode line and disable it then:

` ────── ` (or longer, whatever you see fit)

nilmerg: There's also a bold line char: `━━━━━━━━`

src/FormElement/ScheduleElement/Common/FieldsUtils.php
yhabteab: Even then, my regex still wouldn't be the same. If you're only concerned about the performance, it's only `0.1ms` faster than my version.

src/FormElement/ScheduleElement/FieldsRadio.php
```diff
@@ -0,0 +1,57 @@
...
+                ->registerAttributeCallback('disabled', function () use ($option) {
+                    return $this->getAttributes()->get('disabled')->getValue() || $option->isDisabled();
+                });
+
+            $listItem = HtmlElement::create('li');
+            $listItem->addHtml(
+                $radio,
+                HtmlElement::create('label', [
+                    'for'   => $htmlId,
+                    'class' => $option->getLabelCssClass()
```
nilmerg, line 41: Add `tabindex=-1` as discussed here and remember to disable the outline for the label.

nilmerg (changes requested):

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,314 @@
...
+                    $rule .= "BYDAY=MO,TU,WE,TH,FR;BYSETPOS=$position";
+                } elseif ($runDay === $element::$everyWeekend) {
+                    $rule .= "BYDAY=SA,SU;BYSETPOS=$position";
+                } else {
+                    $rule .= sprintf('BYDAY=%d%s', $position, $runDay);
+                }
+
+                return new RRule($rule);
+        }
+        // Oops!!
```
nilmerg, line 213: What's this? Is this a reason for an error? Then throw an exception.

nilmerg (changes requested):

src/FormElement/ScheduleElement.php
yhabteab: I don't think that will ever happen! But since I don't have a default switch case, PhpStorm renders an error hint.

nilmerg: Ah, well, this hint. I've disabled it for this reason :sweat_smile: 

But even worse is that if it occurs, PHP complains with a cryptic (for the user) error message as the return type must not be null.

Just add a `default` case and throw a `LogicException` there. Then PHPStorm and PHP are happy.

nilmerg (commented):
Not bad. Though, will have a more detailed look tomorrow. One thing I noticed: Upon first visit of a form, the recurrences are shown. If I choose a frequency and then none again, they are gone. Shouldn't the be gone in the first place?

src/FormDecorator/IcingaFormDecorator.php
```diff
@@ -71,6 +72,12 @@ protected function assembleLabel()
         $label = parent::assembleLabel();
         if ($label !== null && ! $this->formElement instanceof FieldsetElement) {
             $label->addWrapper(new HtmlElement('div', Attributes::create(['class' => 'control-label-group'])));
+        } else {
+            $label = new HtmlElement(
+                'div',
+                Attributes::create(['class' => 'control-label-group']),
+                HtmlString::create('&nbsp')
+            );
```
nilmerg, line 80: This must not happen for FieldsetElements! They either have a legend or nothing.

Then you can add the monthly and annually fields by `addElement` in `ScheduleElement`.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,556 @@
...
+        ];
+    }
+
+    public function setDefaultElementDecorator($decorator)
+    {
+        parent::setDefaultElementDecorator($decorator);
+
+        $this->weeklyField->setDefaultElementDecorator($this->getDefaultElementDecorator());
+        $this->monthlyFields->setDefaultElementDecorator($this->getDefaultElementDecorator());
+        $this->annuallyFields->setDefaultElementDecorator($this->getDefaultElementDecorator());
```
nilmerg, line 119: This isn't necessary. The default element decorator is automatically propagated to fieldsets.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,556 @@
...
+            ]);
+        }
+
+        $this->addElement('select', 'frequency', [
+            'required'    => false,
+            'class'       => 'autosubmit',
+            'options'     => [
+                static::NO_REPEAT => $this->translate('None'),
+                'Regular'         => $this->regulars,
+                'Advanced'        => $this->advanced
```
nilmerg, line 389: Please translate the keys!

nilmerg (changes requested):

nilmerg (changes requested):

nilmerg (commented):
So far so good :wink:

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,628 @@
...
+     * @return $this
+     */
+    public function setEnd(DateTime $end): self
+    {
+        $this->end = $end;
+
+        return $this;
+    }
+
+    public function getValue($name = null, $default = null)
```
nilmerg, line 218: I suppose, if `$name` is given, the named element's value should be returned. `getElement($name)->getValue()` is also possible, so why shouldn't this.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    public function setValue($value)
+    {
+        if (isset($value['start']) && $value['start'] instanceof DateTime) {
+            $this->setStart($value['start']);
+        }
+
+        if (isset($value['end'])) {
+            $value['use-end-time'] = 'y';
```
nilmerg, line 343: Not also a call to `setEnd`? You call `setStart` above, that's why I'm asking.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,628 @@
...
+                            $this->setFrequency($frequency->getFrequency());
+                        }
+
+                        break;
+                    default:
+                        $this->setFrequency($frequency->getFrequency());
+                }
+
+                $value['frequency'] = $this->getFrequency();
+                $value['custom-frequency'] = $this->getCustomFrequency();
```
nilmerg, line 411: No matter what we do with `BaseFormElement::init()`, you should maybe avoid calling getters in a setter if possible.

Instead of directly calling `set..` above, you could save these to local variables and then call the setters here where you're storing them in `$value` as well.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,628 @@
...
+            'validators' => [new BetweenValidator($options)]
+        ]);
+
+        $numberSpecifier = HtmlElement::create('div', ['class' => 'number-specifier']);
+        $element = $this->getElement('interval');
+        $element->prependWrapper($numberSpecifier);
+
+        $numberSpecifier->prependHtml(HtmlElement::create('span', null, $this->translate('Every')));
+        $numberSpecifier->addHtml($element);
+        $numberSpecifier->addHtml(HtmlElement::create('span', null, $text));
```
nilmerg, line 589: Please use a `TemplateString` here. Get back to me if you're unsure what to do with it :wink:

nilmerg (changes requested):
`getValue` and `setValue` seem to behave now. Though, what they return and expect, can this be utilized somehow with ipl-scheduler? I mean, is there a counterpart like `RRule::fromFrequency()` but in a more general case that accepts `'none'` and cron expressions as well?

src/FormElement/ScheduleElement.php
yhabteab: Calling `setEnd` does not have too much relevance as `setStart`. The start is not only used by this element but also by the sub elements.

src/FormElement/ScheduleElement.php
yhabteab: Even then, we still have a problem with this.
https://github.com/Icinga/ipl-web/pull/101/commits/37f653cfa4173943427b9766a1921176fdb9102e#diff-3bbb2b8560be734af1a67988cb06e5dad50794bdf9ff2e95228051201215ed0eR353

yhabteab:
> I mean, is there a counterpart like `RRule::fromFrequency()` but in a more general case that accepts `'none'` and cron expressions as well?

No! To me, it doesn't make much sense that the `RRule` class should accept such frequencies.

nilmerg:
Oh sure? So we have [such an abomination](https://github.com/Icinga/icingaweb2-module-x509/blob/provide-job-schedules-widget/application/clicommands/JobsCommand.php#L75-L105) everywhere? :smile: 

The `RRule` class doesn't need to provide such a factory method. I'd imagine a `Frequency::fromConfig()`, though yeah, that's an interface... let's talk in person about this :wink:

src/FormElement/ScheduleElement.php
nilmerg: Okay, but you're not even calling `setEnd` later. Not at all? What's the purpose of `setEnd` then anyway?

yhabteab: To be able to set the end time explicitly from the external, eventually.

nilmerg (approved):
LGTM! Now it's up to me. I will push a test and maybe some CSS updates.

nilmerg (dismissed):
Finished.

Tests fail because https://github.com/Icinga/ipl-scheduler/pull/25 isn't merged yet.

nilmerg:
Please take a look at the tests. They fail for another reason now.

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,592 @@
...
+        if (parent::getValue('use-end-time', 'n') === 'y') {
+            $rule->endAt(parent::getValue('end'));
+        }
+
+        return $rule;
+    }
+
+    public function setValue($value)
+    {
+        $rule = $value['frequency'] ?? null;
```
nilmerg, line 314: If `$value` is not what has been submitted by the browser, it can only be the rule itself already. There's no key `'frequency'` if I populate the element manually.

nilmerg (changes requested):

src/FormElement/ScheduleElement.php
```diff
@@ -0,0 +1,594 @@
...
+     * Get prepared multipart updates
+     *
+     * @param RequestInterface $request
+     *
+     * @return array
+     */
+    public function prepareMultipartUpdate(RequestInterface $request): array
+    {
+        $autoSubmittedBy = $request->getHeader('X-Icinga-AutoSubmittedBy');
+        $pattern = '/^schedule-element\[(weekly-fields|monthly-fields|annually-fields)]'
```
nilmerg, line 568: `^schedule-element` mandates the name of the element. Though, I'd like to have a choice here ;)

nilmerg (changes requested):

src/FormElement/ScheduleElement.php
yhabteab: true!

nilmerg (approved):
LGTM!

Should not be merged until https://github.com/Icinga/ipl-scheduler/pull/25 is merged.

PR #103, opened 2022-10-26
Differentation between equal like unequal and unlike

### LINKED ISSUE #102: Differentiation between equal/unequal and like/unlike in query strings
Equal/Like currently share the same query string representation (`=`) as do unequal and unlike (`!=`).

This makes it impossible to search for literal asterisks.

There should be a distinct representation for like (`~`) and unlike (`!~`).

nilmerg:
resolves #102


sukhwinder33445:
Current changes lead to the following error:

Screenshot 2023-05-31 at 16 12 59


# To Reproduce:

1. In the search bar, add a filter with any operator.
2. Try to replace the operator with `!`.

nilmerg:
Should be fixed

PR #104, opened 2022-11-24
CompatForm: Insert duplicate submit button when more than one is present

TAINCER:
Creates a new `SubmitElement` from the primary submit button of the form with cloned Attributes from the given `FormSubmitElement`. The new submit button is always at the top of the form.
The element has a fixed `primary-submit-btn-duplicate` class, which hides to button, but not the functionality when submitting the form.

### DIFF
--- a/asset/css/primary-submit-btn-duplicate.less
+++ b/asset/css/primary-submit-btn-duplicate.less
@@ -1,0 +1,16 @@
+/**
+  Automatically set CSS class for duplicated submit buttons
+  used for implicit form submission that should be invisible
+  and not take up any space. `display: none` is not an option,
+  because at least Safari will then ignore the element completely
+  when submitting a form.
+ */
+.primary-submit-btn-duplicate {
+  border: 0;
+  height: 0;
+  margin: 0;
+  padding: 0;
+  visibility: hidden;
+  width: 0;
+  position: absolute;
+}
--- a/src/Compat/CompatForm.php
+++ b/src/Compat/CompatForm.php
@@ -2,7 +2,11 @@
 
 namespace ipl\Web\Compat;
 
+use ipl\Html\Contract\FormSubmitElement;
 use ipl\Html\Form;
+use ipl\Html\FormElement\SubmitElement;
+use ipl\Html\HtmlDocument;
+use ipl\Html\HtmlString;
 use ipl\I18n\Translation;
 use ipl\Web\FormDecorator\IcingaFormDecorator;
 
@@ -11,6 +15,27 @@ class CompatForm extends Form
     use Translation;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
+
+    /**
+     * Render the content of the element to HTML
+     *
+     * A duplicate of the primary submit button is being prepended if there is more than one present
+     *
+     * @return string
+     */
+    public function renderContent(): string
+    {
+        if (count($this->submitElements) > 1) {
+            return (new HtmlDocument())
+                ->setHtmlContent(
+                    $this->duplicateSubmitButton($this->submitButton),
+                    new HtmlString(parent::renderContent())
+                )
+                ->render();
+        }
+
+        return parent::renderContent();
+    }
 
     public function hasDefaultElementDecorator()
     {
@@ -21,5 +46,21 @@ class CompatForm extends Form
         $this->setDefaultElementDecorator(new IcingaFormDecorator());
 
         return true;
+    }
+
+    /**
+     * Return a duplicate of the given submit button with the `class` attribute fixed to `primary-submit-btn-duplicate`
+     *
+     * @param FormSubmitElement $originalSubmitButton
+     *
+     * @return SubmitElement
+     */
+    public function duplicateSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
+    {
+        $attributes = (clone $originalSubmitButton->getAttributes())
+            ->set('class', 'primary-submit-btn-duplicate');
+        $attributes->remove('id');
+
+        return new SubmitElement($originalSubmitButton->getName(), $attributes);
     }
 }
--- a/tests/Compat/CompatFormTest.php
+++ b/tests/Compat/CompatFormTest.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace ipl\Tests\Web\Compat;
+
+use ipl\Html\FormElement\SubmitElement;
+use ipl\Tests\Html\TestCase;
+use ipl\Web\Compat\CompatForm;
+
+class CompatFormTest extends TestCase
+{
+    /** @var CompatForm */
+    private $form;
+
+    protected function setUp(): void
+    {
+        $this->form = new CompatForm();
+    }
+
+    public function testDuplicateSubmitButtonApplied(): void
+    {
+        $this->form->addElement('submit', 'submitCreate');
+        $this->form->addElement('submit', 'submitDelete');
+
+        $expected = <<<'HTML'
+    <form class="icinga-form icinga-controls" method="POST">
+      <input class="primary-submit-btn-duplicate" name="submitCreate" type="submit" value="submitCreate"/>
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitCreate" type="submit" value="submitCreate"/>
+      </div>
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitDelete" type="submit" value="submitDelete"/>
+      </div>
+    </form>
+HTML;
+
+        $this->assertHtml($expected, $this->form);
+    }
+
+    public function testDuplicateSubmitButtonOmitted(): void
+    {
+        $this->form->addElement('submit', 'submitCreate');
+
+        $expected = <<<'HTML'
+    <form class="icinga-form icinga-controls" method="POST">
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitCreate" type="submit" value="submitCreate"/>
+      </div>
+    </form>
+HTML;
+
+        $this->assertHtml($expected, $this->form);
+    }
+
+    public function testDuplicateSubmitButtonAddedOnlyOnce(): void
+    {
+        $this->form->addElement('submit', 'submitCreate', ['id' => 'submit_id']);
+        $this->form->addElement('submit', 'submitDelete');
+
+        $expected = <<<'HTML'
+    <form class="icinga-form icinga-controls" method="POST">
+      <input class="primary-submit-btn-duplicate" name="submitCreate" type="submit" value="submitCreate"/>
+      <div class="control-group form-controls">
+        <input id="submit_id" class="btn-primary" name="submitCreate" type="submit" value="submitCreate"/>
+      </div>
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitDelete" type="submit" value="submitDelete"/>
+      </div>
+    </form>
+HTML;
+
+        // Call render twice to ensure that the submit button is only prepended once.
+        $this->form->render();
+        $this->assertHtml($expected, $this->form);
+    }
+
+    public function testDuplicateSubmitButtonRespectsOriginalAttributes(): void
+    {
+        $submitButton = new SubmitElement('test_submit', [
+            'class'          => 'autosubmit',
+            'formnovalidate' => true
+        ]);
+
+        $prefixButton = $this->form->duplicateSubmitButton($submitButton);
+
+        // Name should stay the same
+        $this->assertSame($submitButton->getName(), 'test_submit');
+        $this->assertSame($prefixButton->getName(), 'test_submit');
+
+        // Added attributes should stay the same
+        $this->assertSame($submitButton->getAttributes()->get('formnovalidate')->getValue(), true);
+        $this->assertSame($prefixButton->getAttributes()->get('formnovalidate')->getValue(), true);
+
+        // Class attribute should change to `primary-submit-btn-duplicate`
+        $this->assertSame($submitButton->getAttributes()->get('class')->getValue(), 'autosubmit');
+        $this->assertSame($prefixButton->getAttributes()->get('class')->getValue(), 'primary-submit-btn-duplicate');
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
   },
   "autoload-dev": {
     "psr-4": {
-      "ipl\\Tests\\Web\\": "tests"
+      "ipl\\Tests\\Web\\": "tests",
+      "ipl\\Tests\\Html\\": "vendor/ipl/html/tests"
     }
   },
   "require": {


nilmerg:
What about overriding `setSubmitButton()` with this trait? It's used with child classes of `ipl\Html\Form` after all so it shouldn't pose a problem in the inheritance tree.

I'd like to just use the trait and be done with it. Calling a special method, in this case, is too tedious for me :wink:

TAINCER:
> What about overriding `setSubmitButton()` with this trait? It's used with child classes of `ipl\Html\Form` after all so it shouldn't pose a problem in the inheritance tree.
> 
> I'd like to just use the trait and be done with it. Calling a special method, in this case, is too tedious for me 😉

if I would use the Trait in `setSubmitButton()`, the Prefix Button could be rendered multiple times. The commit for `CompatForm` was still missing, in there I overwrite `renderContent()` with the `createSubmitButton()` method.

src/Compat/CompatForm.php
```diff
@@ -12,6 +14,16 @@ class CompatForm extends Form
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    public function renderContent(): string
+    {
+        if (count($this->submitElements) > 1) {
+            return (new HtmlDocument())
+                ->setHtmlContent($this->createSubmitButton($this->submitButton), new HtmlString(parent::renderContent()));
```
nilmerg, line 21: Isn't this incompatible with the return type hint?

It may be implicitly casted, but I think calling `render()` explicitly is the better approach.

nilmerg:
I see. I forgot that this is only relevant if there are multiple submit buttons and then the set submit button isn't reliable. (even if available)

tests/TestCase.php
```diff
@@ -2,6 +2,34 @@
...
 
+use ipl\Html\ValidHtml;
+use PHPUnit\Util\Xml;
+
+// phpcs:disable
+if (class_exists('PHPUnit_Util_XML')) {
+    // Support older PHPUnit versions
+    class_alias('PHPUnit_Util_XML', 'PHPUnit\\Util\\Xml');
+}
+
```
nilmerg, line 13: Isn't the phpcs:enable instruction required?

tests/TestCase.php
```diff
@@ -2,6 +2,34 @@
...
 abstract class TestCase extends \PHPUnit\Framework\TestCase
 {
+    /** Assert that HTML is equal */
+    protected function assertHtml(string $expectedHtml, ValidHtml $actual): void
```
nilmerg, line 17: While you're at it, please add a third argument `$message` and pass it on to `assertEquals`

tests/TestCase.php
```diff
@@ -2,6 +2,34 @@
...
+
+// phpcs:disable
+if (class_exists('PHPUnit_Util_XML')) {
+    // Support older PHPUnit versions
+    class_alias('PHPUnit_Util_XML', 'PHPUnit\\Util\\Xml');
+}
+
 abstract class TestCase extends \PHPUnit\Framework\TestCase
 {
+    /** Assert that HTML is equal */
```
nilmerg, line 16: I believe you can re-use the docstring from ipl-html, no?

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,79 @@
...
+      <input class="hidePrefixElement" name="submitCreate" type="submit" value="submitCreate"/>
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitDelete" type="submit" value="submitDelete"/>
+      </div>
+      <input name="submitCreate" type="submit" value="submitCreate"/>
+    </form>
+HTML;
+
+        $this->form->render();
+        $this->form->render();
```
nilmerg, line 75: A single call should suffice?

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,79 @@
...
+    private $form;
+
+    protected function setUp(): void
+    {
+        $this->form = new CompatForm();
+    }
+
+    public function testSubmitButtonPrefixApplied(): void
+    {
+        $submitButton = $this->form->createElement('submit', 'submitCreate');
```
nilmerg, line 18: I know why you're not using `addElement` here. But I think the order of submits is not relevant for this test. Please do use it, to reduce complexity.

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,79 @@
...
+      </div>
+    </form>
+HTML;
+
+        $this->assertHtml($expected, $this->form);
+    }
+
+    public function testPrefixSubmitButtonAddedOnlyOnce(): void
+    {
+        $submitButton = $this->form->createElement('submit', 'submitCreate');
```
nilmerg, line 56: Same here, `addElement` please.

src/Common/PrefixSubmitButton.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace ipl\Web\Common;
+
+use ipl\Html\Attribute;
+use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\FormElement\SubmitElement;
+
+trait PrefixSubmitButton
+{
+    protected function createSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
```
nilmerg, line 11: I'm not sure the name of the trait and the method name properly represent their purpose.

For the trait name I think `DuplicateSubmitButton` suits. The same name should be used for the method, as that's what it does. Please document the method as well.

The term *prefix* is not relevant here, only in `CompatForm`.

src/Common/PrefixSubmitButton.php
```diff
@@ -0,0 +1,23 @@
...
+    {
+        $attr = clone $originalSubmitButton->getAttributes();
+        $attr->setAttribute(
+            Attribute::create('class', 'hidePrefixElement')
```
nilmerg, line 15: Consider to also adjust the class name to `submit-btn-duplicate` or similar.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,33 @@
...
 use ipl\Web\FormDecorator\IcingaFormDecorator;
 
 class CompatForm extends Form
 {
     use Translation;
+    use PrefixSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    public function renderContent(): string
```
nilmerg, line 19: Please document here that this injects a duplicate of the submit button at the start.

nilmerg (changes requested):

tests/TestCase.php
TAINCER: The [`asertHtml`](https://github.com/Icinga/ipl-html/blob/master/tests/TestCase.php#L18-L23) from ipl-html did not have types for everything. I added the missing Types, so the PHPDocs would become redundant because they are already present in the Method Signature. I can add them back If we want them regardless, the actual comment stayed the same.

tests/TestCase.php
TAINCER: The test seems to work with phpcs:enable. I removed the whole line, since phpcs:enable should be set by default.
This is also from the Original method in ipl-html. Maybe it caused Problems with older PHP versions.

tests/TestCase.php
```diff
@@ -2,6 +2,41 @@
...
+    /**
+     * Assert that HTML is equal
+     *
+     * @param string    $expectedHtml
+     * @param ValidHtml $actual
+     * @param String    $message
+     *
+     * @return void
+     */
+    protected function assertHtml(string $expectedHtml, ValidHtml $actual, String $message = ''): void
```
nilmerg, line 24: The `string` type name always all lowercase please.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,38 @@
...
 
 class CompatForm extends Form
 {
     use Translation;
+    use DuplicateSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * A duplicate of the Primary Submit Button is being prepended, if there are more than one present.
```
nilmerg, line 20: You're overriding the base documentation block. Make sure to include the short message here again.

nilmerg (changes requested):
Please also add a another test case to `CompatFormTest` which verifies that an explicit call to `setSubmitButton()` duplicates said submit button, and not the first registered one.

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace ipl\Web\Common;
+
+use ipl\Html\Attribute;
+use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\FormElement\SubmitElement;
+
+trait DuplicateSubmitButton
+{
+    /** Creates a Duplicate Submit Element with an added `class` attribute */
```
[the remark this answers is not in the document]
nilmerg: nvm.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,34 @@
...
 use ipl\Web\FormDecorator\IcingaFormDecorator;
 
 class CompatForm extends Form
 {
     use Translation;
+    use DuplicateSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    // A duplicate of the Primary Submit Button is being prepended, if there are more than one present.
```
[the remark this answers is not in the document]
nilmerg: nvm.

tests/TestCase.php
nilmerg: Oh well, nevermind. Please extend `ipl\Tests\Html\TestCase` in `CompatFormTest`. Then this isn't required at all.

nilmerg (approved):
LGTM!

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
```
nilmerg, line 2: Please add a namespace

tests/common/DuplicateSubmitButtonTest.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
```
nilmerg, line 2: Please add a namespace. And make the directory name use an uppercase `C`.

nilmerg (changes requested):
Please squash your commits to the following:

* One for the trait + less file
* One for the composer.json
* One for the CompatFormTest
* One for the DuplicateSubmitButttonTest + TestCase

tests/TestCase.php
```diff
@@ -4,4 +4,5 @@
 
 abstract class TestCase extends \PHPUnit\Framework\TestCase
 {
+
```
lippserd, line 7: Please remove.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,39 @@
...
 class CompatForm extends Form
 {
     use Translation;
+    use DuplicateSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * Render the content of the element to HTML
+     * A duplicate of the Primary Submit Button is being prepended, if there are more than one present.
```
lippserd, line 21: Please correct spelling and capitalization.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,39 @@
...
 
 class CompatForm extends Form
 {
     use Translation;
+    use DuplicateSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * Render the content of the element to HTML
```
lippserd, line 20: Shouldn't there be a newline after this?

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Creates a Duplicate Submit Element with an added `class` attribute
+     *
+     * @param FormSubmitElement $originalSubmitButton
+     *
+     * @return SubmitElement
+     */
+    protected function duplicateSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
+    {
+        $attr = clone $originalSubmitButton->getAttributes();
```
lippserd, line 20: Aren't we cloning the element because of a possible wrapper or because the cloning PR hasn't been merged yet?

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use ipl\Html\Attribute;
+use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\FormElement\SubmitElement;
+
+trait DuplicateSubmitButton
```
lippserd, line 9: Please describe the actual use case of this trait.

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Creates a Duplicate Submit Element with an added `class` attribute
```
lippserd, line 12: Please correct spelling and capitalization.

lippserd (changes requested):

src/Common/DuplicateSubmitButton.php
TAINCER: Primarily because the cloning PR isn't merged yet. But I also haven't tested if there could be an unwanted wrapper if I would clone the whole element.

src/Compat/CompatForm.php
```diff
@@ -3,15 +3,39 @@
...
 
 class CompatForm extends Form
 {
     use Translation;
+    use DuplicateSubmitButton;
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * Render the content of the element to HTML.
```
lippserd, line 20: Remove the period at the end of the and add a newline.

tests/Common/DuplicateSubmitButtonTest.php
```diff
@@ -0,0 +1,29 @@
...
+            'class' => 'autosubmit'
+        ]);
+
+        $prefixButton = $this->duplicateSubmitButton($submitButton);
+
+        // Name should stay the same
+        $this->assertSame($submitButton->getName(), 'test_submit');
+        $this->assertSame($prefixButton->getName(), 'test_submit');
+
+        // Class Attribute should change to `hideElement`
```
lippserd, line 25: `attribute`

asset/css/prefix-button.less
```diff
@@ -0,0 +1,9 @@
+.submit-btn-duplicate {
```
lippserd, line 1: I would call it `primary-submit-btn-duplicate`. And please fix the filename as well.

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,33 @@
...
+namespace ipl\Web\Common;
+
+use ipl\Html\Attribute;
+use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\FormElement\SubmitElement;
+
+/**
+ * Provides the `duplicateSubmitButton` method, used to force the primary submit element in forms
+ */
+trait DuplicateSubmitButton
```
lippserd, line 12: In my opinion, this trait is no longer useful. It is called `DuplicateSubmitButton`, although its intent is to always clone the primary submit button, and it also hard codes the CSS class to use. I opt to remove the trait and just clone directly in the form. Please keep that in the method as it is easier to test.

src/Common/DuplicateSubmitButton.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Returns a duplicate submit element with an added `class` attribute
+     *
+     * @param FormSubmitElement $originalSubmitButton
+     *
+     * @return SubmitElement
+     */
+    protected function duplicateSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
+    {
+        $attr = clone $originalSubmitButton->getAttributes();
```
lippserd, line 23: Please call the variable `$attributes`.

lippserd (changes requested):

src/Compat/CompatForm.php
```diff
@@ -12,6 +17,27 @@ class CompatForm extends Form
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * Render the content of the element to HTML
+     *
+     * A duplicate of the primary submit button is being prepended if there is more than one present
```
lippserd, line 23: Dot at the end of the line missing.

src/Compat/CompatForm.php
```diff
@@ -22,4 +48,24 @@ public function hasDefaultElementDecorator()
 
         return true;
     }
+
+    /**
+     * Returns a duplicate submit element with an added `class` attribute
```
lippserd, line 53: This sentence needs to be corrected several times:

1. `Return` vs `Returns`.
2. It doesn't  add a `class` attributes, it overrides it with a specific value.
3. We're talking about `submit button` not `submit element` throughout this PR.

Consider rephrasing to:

> Return a duplicate of the given submit button with the `class` attribute fixed to `primary-submit-btn-duplicate`

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    /** @var CompatForm */
+    private $form;
+
+    protected function setUp(): void
+    {
+        $this->form = new CompatForm();
+    }
+
+    public function testSubmitButtonPrefixApplied(): void
```
lippserd, line 19: You're still using `prefix` in method names.

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'class' => 'autosubmit'
+        ]);
+
+        $prefixButton = $this->form->duplicateSubmitButton($submitButton);
+
+        // Name should stay the same
+        $this->assertSame($submitButton->getName(), 'test_submit');
+        $this->assertSame($prefixButton->getName(), 'test_submit');
+
+        // Class attribute should change to `hideElement`
```
lippserd, line 88: `hideElement` is no longer valid.

asset/css/primary-submit-btn-duplicate.less
```diff
@@ -0,0 +1,9 @@
+.primary-submit-btn-duplicate {
```
lippserd, line 1: I would add a description here, something like:

> Automatically set CSS class for duplicated submit buttons used for implicit form submission that should be invisible and not take up any space. `display: none` is not an option, because at least Safari will then ignore the element completely.

lippserd (changes requested):

lippserd (changes requested):
PR title and description have to be adjusted as well.

nilmerg (dismissed):

src/Compat/CompatForm.php
```diff
@@ -22,4 +47,28 @@ public function hasDefaultElementDecorator()
...
+    /**
+     * Return a duplicate of the given submit button with the `class` attribute fixed to `primary-submit-btn-duplicate`
+     *
+     * @param FormSubmitElement $originalSubmitButton
+     *
+     * @return SubmitElement
+     */
+    public function duplicateSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
+    {
+        $attributes = array_reduce(
```
nilmerg, line 60: I can't imagine a valid reason for using `array_reduce` here.

Though, I guess the intention was to have the same attributes on the duplicate as on the original? This fails miserably because a) `array_reduce` returns a single result if used correctly and b) the current incorrect usage returns an empty array, always.

Though, even if corrected, what's the actual reason for transferring the attributes anyway? The duplicate just needs the correct name and value. Anything else is not required, or is it? A problem I didn't noticed previously is also that the ID attribute is duplicated in this case, but duplicate IDs are invalid in HTML.

nilmerg (changes requested):

src/Compat/CompatForm.php
lippserd: The submit button to be cloned can have attributes and callbacks that must also be present on the clone, such as `formaction`, `formmethod` and `formnovalidate`, so simply not using the original attributes is not an option.

With Icinga/ipl-html#110 merged, I suggest somethin like the following code here:

```php
// Until proper cloning (ref here) is implemented ...
$attributes = clone $originalSubmitButton->getAttributes();
$submit = new SubmitElement($originalSubmitButton->getName(), $attributes);
$attributes->registerAttributeCallback('name', [$submit, 'getNameAttribute']);

Comment (TAINCER) [reply src/Compat/CompatForm.php:60]: I've added a test for additional attributes to be carried over. I also fixed the `array_reduce` - I was missing a return.

Regarding the ID, how do you think this should be handled? Either an appended name or a unique name would work.

Comment (nilmerg) [reply src/Compat/CompatForm.php:60]: Attribute cloning is now merged.

File: src/Compat/CompatForm.php [hunk 34]
```diff
@@ -22,4 +47,25 @@ public function hasDefaultElementDecorator()
...
+     * @return SubmitElement
+     */
+    public function duplicateSubmitButton(FormSubmitElement $originalSubmitButton): SubmitElement
+    {
+        $attributes = (clone $originalSubmitButton->getAttributes())
+            ->set('class', 'primary-submit-btn-duplicate');
+        $attributes->remove('id');
+
+        $submit = (new SubmitElement($originalSubmitButton->getName(), $attributes))
+            ->setValue($originalSubmitButton->getValue());
```
lippserd, line 65: You don't have to set the value here as it has no effect. The original attributes have a callback for that.

src/Compat/CompatForm.php
```diff
@@ -12,6 +16,27 @@ class CompatForm extends Form
 
     protected $defaultAttributes = ['class' => 'icinga-form icinga-controls'];
 
+    /**
+     * Render the content of the element to HTML
+     *
+     * A duplicate of the primary submit button is being prepended if there is more than one present.
```
lippserd, line 22: Please remove the dot at the end.

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,97 @@
...
+      </div>
+    </form>
+HTML;
+
+        $this->form->render();
+
+        $this->assertHtml($expected, $this->form);
+    }
+
+    public function testCreatingDuplicateSubmitButton(): void
```
lippserd, line 76: `testDuplicateSubmitButtonRespectsOriginalAttributes`

lippserd (changes requested):

tests/Compat/CompatFormTest.php
```diff
@@ -0,0 +1,97 @@
...
+      <div class="control-group form-controls">
+        <input id="submit_id" class="btn-primary" name="submitCreate" type="submit" value="submitCreate"/>
+      </div>
+      <div class="control-group form-controls">
+        <input class="btn-primary" name="submitDelete" type="submit" value="submitDelete"/>
+      </div>
+    </form>
+HTML;
+
+        $this->form->render();
```
lippserd, line 71: Why the render call?

TAINCER: To render it once, and again in the `assertHtml()`. To make sure it only prepends one submit button.

lippserd: I would add  something like this as comment before the render call and remove the newline before `asserHtml()` as this block is the test case, i.e.

```php
// Submit button is duplicated in render(), so we have to call it twice for the test.
render()
assert()

lippserd (approved):

nilmerg (approved):

PR #108, opened 2023-01-13
Introduce `BaseScheduleForm` class

yhabteab:


nilmerg:
Changes are now incorporated in #101

PR #117, opened 2023-02-27
CompatFormTest: Always add `control-label-group` group

TAINCER:
Fixes missing empty groups introduced by a commit to [IcingaFormDecorator.php](https://github.com/Icinga/ipl-web/commit/e5e8c2011ea11923c7fe98a2dddd4d2214d37470).

### DIFF
--- a/tests/Compat/CompatFormTest.php
+++ b/tests/Compat/CompatFormTest.php
@@ -25,9 +25,11 @@ class CompatFormTest extends TestCase
     <form class="icinga-form icinga-controls" method="POST">
       <input class="primary-submit-btn-duplicate" name="submitCreate" type="submit" value="submitCreate"/>
       <div class="control-group form-controls">
+        <div class="control-label-group">&amp;nbsp</div>
         <input class="btn-primary" name="submitCreate" type="submit" value="submitCreate"/>
       </div>
       <div class="control-group form-controls">
+        <div class="control-label-group">&amp;nbsp</div>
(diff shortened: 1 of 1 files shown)


nilmerg:
We should think about not rendering this empty div for form controls. They never had a label to begin with and the referenced commit targeted only normal elements.

TAINCER:
> We should think about not rendering this empty div for form controls. They never had a label, and the referenced commit targeted only normal elements.

I don't know what "normal elements" mean in this context, but rendering empty divs also seems wasteful to me. I talked with @yhabteab about this, but if this is not intentional, maybe only render the wrapper if there is a label or at least one element is present.

Something like (not tested):
```php
    protected function assembleLabel()
    {
        $label = parent::assembleLabel();
        if (! $this->formElement instanceof FieldsetElement) {
            if ($label !== null || ! parent::isEmpty()) {
                $label->addWrapper(new HtmlElement('div', Attributes::create(['class' => 'control-label-group'])));
            }
        }

        return $label;
    }
```

yhabteab:
You just have to change this condition as follows:
https://github.com/Icinga/ipl-web/blob/e0b4e25f11b3d03a2cebc8db986d2c507c3eb0dc/src/FormDecorator/IcingaFormDecorator.php#L76

```php
        if (! $this->formElement instanceof FieldsetElement && ! $this->formElement instanceof SubmitElement) {
```

PR #118, opened 2023-02-27
IcingaFormDecorator: Don't wrap `SubmitButtonElements` in an empty `control-label-group` div

TAINCER:
closes #117

### DIFF
--- a/src/FormDecorator/IcingaFormDecorator.php
+++ b/src/FormDecorator/IcingaFormDecorator.php
@@ -76,7 +76,7 @@ class IcingaFormDecorator extends DivDecorator
         if (! $this->formElement instanceof FieldsetElement) {
             if ($label !== null) {
                 $label->addWrapper(new HtmlElement('div', Attributes::create(['class' => 'control-label-group'])));
-            } else {
+            } elseif (! $this->formElement instanceof FormSubmitElement) {
                 $label = new HtmlElement(
                     'div',
                     Attributes::create(['class' => 'control-label-group']),
@@ -96,9 +96,9 @@ class IcingaFormDecorator extends DivDecorator
 
         if (($description = $this->formElement->getDescription()) !== null) {
             $iconAttributes = [
-                'class'         => 'control-info',
-                'role'          => 'image',
-                'title'         => $description
+                'class' => 'control-info',
+                'role'  => 'image',
+                'title' => $description
             ];
 
             $describedBy = null;


src/FormDecorator/IcingaFormDecorator.php
```diff
@@ -73,16 +75,23 @@ protected function createCheckbox(CheckboxElement $checkbox)
...
-                    'div',
-                    Attributes::create(['class' => 'control-label-group']),
-                    HtmlString::create('&nbsp')
-                );
-            }
+
+        if (
+            $this->formElement instanceof FieldsetElement ||
+            $this->formElement instanceof SubmitElement ||
+            $this->formElement instanceof SubmitButtonElement
```
nilmerg, line 82: There's the `FormSubmitElement` interface. Use it instead and the `if` doesn't need to be disassembled. You're introducing way to much noise in the change history with this litte change.

nilmerg (changes requested):

lippserd (commented):
If I understand the change correctly, it does the following:

* It no longer creates an empty `div` for the **label** of submit elements. The submit element will still be wrapped.
* A wrapper `div` for the `legend` or an empty `div` is now created for `FieldsetElement`. 

If this is correct, please adjust commit message and description as well as PR title and description. As the current state does not reflect the actual change.

TAINCER:
> * It no longer creates an empty `div` for the **label** of submit elements. The submit element will still be wrapped.
Yes, primarily to fix the failing [CompatFormTest](https://github.com/Icinga/ipl-web/actions/runs/4254446866/jobs/7400709057#step:5:21).

> * A wrapper `div` for the `legend` or an empty `div` is now created for `FieldsetElement`.
No, I misunderstood the previous change request, this should remain as it was before.

lippserd:
> > * A wrapper `div` for the `legend` or an empty `div` is now created for `FieldsetElement`.
> 
> No, I misunderstood the previous change request, this should remain as it was before.

So this PR is not ready to review yet, correct?

nilmerg (approved):

TAINCER:
> > > * A wrapper `div` for the `legend` or an empty `div` is now created for `FieldsetElement`.
> > 
> > 
> > No, I misunderstood the previous change request, this should remain as it was before.
> 
> So this PR is not ready to review yet, correct?

It is now, sorry. Also, this should not create an empty div for the `SubmitButtonElements` and not remove them entirely. I updated the title for that.

lippserd (approved):

PR #122, opened 2023-03-07
Enhance schedule element fields validation

yhabteab:


src/FormElement/ScheduleElement.php
```diff
@@ -565,6 +565,17 @@ private function assembleCommonElements(): void
      */
     public function prepareMultipartUpdate(RequestInterface $request): array
     {
+        $hasErrors = $this->weeklyField->getPopulatedValue('has_errors') === 'y';
```
lippserd, line 568: Why do we have transfer `has_errors`? Is there really no other way? Looks fragile.

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -86,6 +86,30 @@ public function getSelectedDays(): array
...
 
+    public function validate()
+    {
+        parent::validate();
+        if (! $this->isValid || $this->getPopulatedValue('runsOn', static::RUNS_EACH) === static::RUNS_ONTHE) {
+            return $this;
+        }
+
+        $this->isValid = false;
+        foreach (range(1, $this->availableFields) as $day) {
```
lippserd, line 97: Can't this be done via `hasValue()`?

tests/ScheduleElementTest.php
```diff
@@ -20,14 +20,13 @@ protected function setUp(): void
     protected function assembleElement(array $options): ScheduleElement
     {
         $element = new ScheduleElement('test', $options);
+        $element->render(); // Forces also assembly of any content
```
lippserd, line 23: Why `render()` instead of `ensureAssembled()`?

yhabteab: Because `render()` is calling `ensureAssembled()` automatically before rendering other elements that provide `renderWrapped()` & `renderUnwrapped()` implementation.

src/FormElement/ScheduleElement.php
yhabteab: In the normal case we only update the content partially i.e. just the `Recurrence` widget. However, if an error was rendered before for some reason, this message will be displayed forever even if the user has already chosen a valid option in between.

I would be willing to accept better suggestions on how to make this work in a much nicer way.

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -119,8 +142,10 @@ protected function assemble()
             $listItems->addHtml($listItem);
         }
 
-        if (! $foundCheckedDay) {
-            $this->getElement($this->default)->setChecked(true);
+        if (! $foundCheckedDay && $this->getPopulatedValue($this->default) !== 'n') {
+            $this->getElement($this->default)
+                ->setValue('y')
+                ->setChecked(true);
```
nilmerg, line 148: This is not required if you correctly define the initial value of each checkbox.

src/FormElement/ScheduleElement/WeeklyFields.php
```diff
@@ -95,6 +95,29 @@ public function loadWeekDays(array $days): array
         return $values;
     }
 
+    public function validate()
```
nilmerg, line 98: Isn't it possible to achieve this by a default validator? I don't quite like that you're overriding `validate()`.

src/FormElement/ScheduleElement.php
```diff
@@ -495,6 +496,43 @@ protected function assemble()
                     'valid'     => function (): bool {
                         return $this->isValid();
                     },
+                    'validators' => [
```
nilmerg, line 499: I don't quite like the usage of a validator here. You're essentially mis-using it to define the element's content.

Can you use the `onRegister()` method of the element? This way you can remember the schedule element as parent and also replace the `valid` callback.

The parent is then used to decide what to do in `assemble()`.

nilmerg (changes requested):

src/FormElement/ScheduleElement/WeeklyFields.php
yhabteab: I enable the checkbox after all elements are registered/added, how would the initial value help here. If then this would have to be fixed in `CheckboxElement`, so that `setChecked()` call also sets the value to `y`.

nilmerg (changes requested):

src/FormElement/ScheduleElement/MonthlyFields.php
```diff
@@ -166,4 +162,30 @@ protected function registerAttributeCallbacks(Attributes $attributes)
             ->registerAttributeCallback('availableFields', null, [$this, 'setAvailableFields'])
             ->registerAttributeCallback('protector', null, [$this, 'setIdProtector']);
     }
+
+    protected function addDefaultValidators(ValidatorChain $chain): void
+    {
+        $chain->add(
+            new CallbackValidator(function ($_, CallbackValidator $validator): bool {
+                if ($this->getPopulatedValue('runsOn', static::RUNS_EACH) !== static::RUNS_EACH) {
```
nilmerg, line 170: Why `getPopulatedValue` here and `getValue` below?

src/FormElement/ScheduleElement/MonthlyFields.php
nilmerg: No, otherwise it's not validated partially. (now)

src/FormElement/ScheduleElement/MonthlyFields.php
yhabteab: Makes no difference!

nilmerg: Then please use `getValue` exclusively.

tests/ScheduleElementTest.php
nilmerg: Not sure why you need to change this. It doesn't make a difference (anymore?) if I run the tests.

nilmerg (changes requested):
If I enter an invalid value into the interval field, an exception is shown. This should be validated correctly.

yhabteab:
> If I enter an invalid value into the interval field, an exception is shown. This should be validated correctly.

What do you mean by an "invalid value"? Actually, this should never happen.

nilmerg:
Zero. A negative number.

yhabteab:
Then something must be broken with the validation in `ipl-html`. There is a `BetweenValidator()` for exactly such use cases.

nilmerg:
Yeah, there is, but you access the value without verifying that it is valid:

```
Interval must be a positive integer

#0 /usr/share/icinga-php/vendor/vendor/simshaun/recurr/src/Recurr/Rule.php(372): Recurr\Rule->setInterval(Integer)
#1 /usr/share/icinga-php/vendor/vendor/simshaun/recurr/src/Recurr/Rule.php(275): Recurr\Rule->loadFromArray(Array)
#2 /usr/share/icinga-php/vendor/vendor/simshaun/recurr/src/Recurr/Rule.php(214): Recurr\Rule->loadFromString(String)
#3 /usr/local/src/ipl-scheduler/src/RRule.php(70): Recurr\Rule->__construct(String)
#4 /usr/local/src/ipl-web/src/FormElement/ScheduleElement.php(236): ipl\Scheduler\RRule->__construct(String)
#5 /usr/local/src/ipl-html/src/FormElement/BaseFormElement.php(269): ipl\Web\FormElement\ScheduleElement->getValue()
#6 /usr/local/src/ipl-html/src/FormElement/FieldsetElement.php(42): ipl\Html\FormElement\BaseFormElement->validate()
#7 /usr/local/src/ipl-html/src/Form.php(331): ipl\Html\FormElement\FieldsetElement->validate()
#8 /usr/local/src/ipl-html/src/Form.php(237): ipl\Html\Form->validatePartial()
#9 /usr/share/icingaweb2-modules/icingadb/application/controllers/TestController.php(343): ipl\Html\Form->handleRequest(Object(GuzzleHttp\Psr7\ServerRequest))
#10 /icingaweb2/library/vendor/Zend/Controller/Action.php(507): Icinga\Module\Icingadb\Controllers\TestController->scheduleElementTestAction()
#11 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch(String)
#12 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#13 /icingaweb2/library/Icinga/Application/Web.php(290): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#14 /icingaweb2/library/Icinga/Application/webrouter.php(105): Icinga\Application\Web->dispatch()
#15 /icingaweb2/public/index.php(4): require_once(String)
#16 {main}
```

nilmerg (approved):

nilmerg:
Due to upcoming changes in ipl-html, you'll also need to change this:

* Don't try to assemble a recurrency rule in `getValue()` if `$this->hasBeenValidated()` is false, call the base implementation instead.
* Override `hasValue()`. You'll return a non-array in `getValue()` and the base implementation does not expect that anymore.
* Call `ensureAssembled` in `prepareMultipartUpdate()`, otherwise the preview element is missing.

Perform your own tests as well please. These things are just what my own tests revealed.

yhabteab:
The unit tests require the following PR to pass.
- https://github.com/Icinga/ipl-html/pull/121

nilmerg (changes requested):
Please squash your commits. The first now introduces `hasValue` and the second removes it.

nilmerg (approved):

lippserd (changes requested):
I have no clue what I'm reviewing. Neither commit/PR title/description describes what's going here.

yhabteab:
> I have no clue what I'm reviewing. Neither commit/PR title/description describes what's going here.

You can now review the commits individually.

lippserd (approved):

PR #127, opened 2023-03-21
CompatForm: Register form element namespace

nilmerg:
blocked by https://github.com/Icinga/ipl-html/pull/124


nilmerg:
Nope. Not in the constructor. But `ensureDefaultElementLoaderRegistered` is not an option as `$defaultElementLoaderRegistered` is private -.-. I won't change it's visibility **now**, though.

PR #131, opened 2023-03-30
font: Add `icinga` glyph to Icinga Icons

flourish86:
Adds an icinga logo glyph and makes it available with `icon iicon-icinga ` css class.


nilmerg (approved):
Looks nice!

PR #133, opened 2023-04-04
Do not autosubmit start with `NO_REPEAT` freq

### LINKED ISSUE #130: Changing start time by keyboard without having a frequency chosen fails badly
With the new schedule element, don't choose a frequency but change the start time by keyboard. This error is shown:

```
Can't get element 'schedule-recurrences'. Element does not exist

#0 /usr/local/src/ipl-web/src/FormElement/ScheduleElement.php(602): ipl\Html\FormElement\FieldsetElement->getElement(String)
```

yhabteab:
fixes #130

### DIFF
--- a/src/FormElement/ScheduleElement.php
+++ b/src/FormElement/ScheduleElement.php
@@ -125,7 +125,7 @@ class ScheduleElement extends FieldsetElement
      */
     public function getFrequency(): string
     {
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
The start date does not auto submit at all anymore now. Not even with a frequency set.

nilmerg (approved):

PR #136, opened 2023-04-11
Fix duplicate submit button not triggering `getPressedSubmitElement()`

### LINKED ISSUE #132: Submitting a form by pushing Enter does not work
That's because only the attributes are cloned and the label that has been set on the original submit button is a property of the element.

TAINCER:
The duplicated submit button does not have the same label as the original submit element. Thus the check in [getPressedSubmitElement()](https://github.com/Icinga/ipl-html/blob/8740b9efd42607e1b03a792897a9d3d51bc01e7f/src/Form.php#L148) can never be `true`.

The submit button check [hasBeenPressed()](https://github.com/Icinga/ipl-html/blob/8740b9efd42607e1b03a792897a9d3d51bc01e7f/src/FormElement/SubmitElement.php#L43) expects the label in the if statement, currently the name is given.

fixes #132

### DIFF
--- a/tests/Compat/CompatFormTest.php
+++ b/tests/Compat/CompatFormTest.php
@@ -36,6 +36,62 @@ HTML;
         $this->assertHtml($expected, $this->form);
     }
 
+    public function testSubmitElementDuplication(): void
+    {
+        $this->form->addElement('submit', 'submit', [
+            'label' => 'Submit label',
+            'class' => 'btn-primary'
+        ]);
+        $this->form->addElement('submit', 'delete', [
+            'label' => 'Delete label',
+            'class' => 'btn-danger'
+        ]);
+        $this->form->setSubmitButton($this->form->getElement('submit'));
+
+        $expected = <<<'HTML'
+    <form class="icinga-form icinga-controls" method="POST">
+      <input class="primary-submit-btn-duplicate" name="submit" type="submit" value="Submit label"/>
+      <div class="control-group form-controls">
+        <input class="btn-primary btn-primary" name="submit" type="submit" value="Submit label"/>
+      </div>
+      <div class="control-group form-controls">
+        <input class="btn-danger btn-primary" name="delete" type="submit" value="Delete label"/>
+      </div>
+    </form>
+HTML;
+
+        $this->assertHtml($expected, $this->form);
+    }
+
+
+    public function testSubmitButtonElementDuplication(): void
+    {
+        $this->form->addElement('submitButton', 'submit', [
+            'label' => 'Submit label',
+            'class' => 'btn-primary',
+            'value' => 'submit_value'
+        ]);
+        $this->form->addElement('submitButton', 'delete', [
+            'label' => 'Delete label',
+            'class' => 'btn-danger'
+        ]);
+        $this->form->setSubmitButton($this->form->getElement('submit'));
+
+        $expected = <<<'HTML'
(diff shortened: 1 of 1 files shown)


src/Compat/CompatForm.php
```diff
@@ -61,6 +61,14 @@ public function duplicateSubmitButton(FormSubmitElement $originalSubmitButton):
             ->set('class', 'primary-submit-btn-duplicate');
         $attributes->remove('id');
 
-        return new SubmitElement($originalSubmitButton->getName(), $attributes);
+        $newSubmitButton = new SubmitElement($originalSubmitButton->getName(), $attributes);
+
+        if ($originalSubmitButton instanceof SubmitElement) {
+            $newSubmitButton->setLabel($originalSubmitButton->getButtonLabel());
+        } else {
+            $newSubmitButton->setLabel($originalSubmitButton->getLabel());
```
nilmerg, line 69: The label of SubmitButtonElements have no purpose in this regard, they exist only for display reasons. It's the submit value that needs to be kept in this case.

nilmerg (changes requested):
Also, please fix/extend the tests. They're also flawed.

src/Compat/CompatForm.php
```diff
@@ -53,14 +55,30 @@ public function hasDefaultElementDecorator()
      *
      * @param FormSubmitElement $originalSubmitButton
      *
-     * @return SubmitElement
+     * @return SubmitButtonElement
```
nilmerg, line 58: This should be `FormSubmitElement`

src/Compat/CompatForm.php
```diff
@@ -53,14 +55,30 @@ public function hasDefaultElementDecorator()
...
+        $attributes->remove('type');
 
-        return new SubmitElement($originalSubmitButton->getName(), $attributes);
+        if ($originalSubmitButton instanceof SubmitElement) {
+            $newSubmitButton = new SubmitElement($originalSubmitButton->getName(), $attributes);
+            $newSubmitButton->setLabel($originalSubmitButton->getButtonLabel());
+            return $newSubmitButton;
+        } elseif ($originalSubmitButton instanceof SubmitButtonElement) {
+            $newSubmitButton = new SubmitButtonElement($originalSubmitButton->getName(), $attributes);
+            $newSubmitButton->setLabel($originalSubmitButton->getSubmitValue());
```
nilmerg, line 74: That should be of no interest, it's hidden after all

src/Compat/CompatForm.php
```diff
@@ -53,14 +55,30 @@ public function hasDefaultElementDecorator()
...
             ->set('class', 'primary-submit-btn-duplicate');
         $attributes->remove('id');
+        // Remove to avoid `type="submit submit"` in SubmitButtonElement
+        $attributes->remove('type');
 
-        return new SubmitElement($originalSubmitButton->getName(), $attributes);
+        if ($originalSubmitButton instanceof SubmitElement) {
+            $newSubmitButton = new SubmitElement($originalSubmitButton->getName(), $attributes);
+            $newSubmitButton->setLabel($originalSubmitButton->getButtonLabel());
+            return $newSubmitButton;
```
nilmerg, line 71: empty line before `return` please

nilmerg (changes requested):

nilmerg (approved):

PR #137, opened 2023-04-12
Make icon style adjustable

sukhwinder33445:
### Required in:
https://github.com/Icinga/noma-web/pull/23

### DIFF
--- a/src/Widget/IcingaIcon.php
+++ b/src/Widget/IcingaIcon.php
@@ -9,16 +9,16 @@ class IcingaIcon extends Icon
     protected $style = '';
 
     /**
-     * Create a icon element
+     * Create an icon element
      *
-     * Creates a icon element from the given name and HTML attributes. The icon element's tag will be <i>. The given
+     * Creates an icon element from the given name and HTML attributes. The icon element's tag will be <i>. The given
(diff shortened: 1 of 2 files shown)


nilmerg (changes requested):
I really wonder why this was working for you when you pushed it. For me it breaks the entire layout.

src/Widget/IcingaIcon.php
```diff
@@ -6,17 +6,19 @@
 
 class IcingaIcon extends Icon
 {
+    protected $defaultStyle;
```
nilmerg, line 9: I'd rather set `$style` to the empty string by default.

src/Widget/Icon.php
```diff
@@ -12,21 +12,60 @@ class Icon extends BaseHtmlElement
...
+    public function setStyle(string $style): self
+    {
+        $this->style = $style;
+
+        return $this;
+    }
+
+    protected function assemble()
+    {
+        $iconStyle = $this->getStyle() ?? $this->defaultStyle;
```
nilmerg, line 65: Please put the fallback into the getter directly.

nilmerg: And don't forget to adjust the return type then!

nilmerg (changes requested):

nilmerg (approved):

PR #148, opened 2023-05-17
Fix filtering when '*' is present in the value string

raviks789:
Treat '*' as wildcard only when it is present in starting or end position of the value string.

fixes Icinga/icingadb-web#693


nilmerg:
We cannot do it this way. A wildcard must be valid anywhere in a value.

This should be addressed in Icinga DB Web with the help of #103.

PR #151, opened 2023-05-25
Add copy to clipboard button widget

raviks789:
The button copies the text from the specified target id when mentioned otherwise it copies the text from the parent element.

ref Icinga/icingadb-web#677

### DIFF
--- a/asset/css/copy-to-clipboard.less
+++ b/asset/css/copy-to-clipboard.less
@@ -8,22 +8,26 @@
   }
 
   &:hover {
-    .copy-to-clipboard {
+    .copy-to-clipboard.active {
       opacity: 0.8;
     }
   }
 }
 
 .copy-to-clipboard {
-  appearance: none;
+  .appearance(none);
   border: none;
   opacity: 0;
   background: none;
   padding: 0.25em 0.5em;
+  cursor: default;
+}
 
+.active {
+  cursor: pointer;
   &:hover, &:focus {
     opacity: 0.8;
-    color: @control-color;
+    color: var(--control-color, @control-color);;
     outline-offset: -2px;
   }
 }
--- a/asset/css/copy-to-clipboard.less
+++ b/asset/css/copy-to-clipboard.less
@@ -1,0 +1,35 @@
+.clipboard-wrapper {
+  position: relative;
+  > .copy-to-clipboard {
+    margin: 0.1em 0.1em 0 0;
+    position: absolute;
+    right: 0;
+    top: 0;
+  }
+
+  &:hover {
+    .copy-to-clipboard {
+      opacity: 0.8;
+    }
+  }
+}
+
+.copy-to-clipboard {
+  appearance: none;
+  border: none;
+  opacity: 0;
+  background: none;
+  padding: 0.25em 0.5em;
+
+  &:hover, &:focus {
+    opacity: 0.8;
+    color: @control-color;
+    outline-offset: -2px;
+  }
+}
+
+.copy-to-clipboard.copied,
+.clipboard-wrapper:hover .copy-to-clipboard.copied {
+  opacity: 0;
+  transition: opacity 1s linear;
+}
--- a/asset/css/copy-to-clipboard.less
+++ b/asset/css/copy-to-clipboard.less
@@ -21,19 +21,21 @@
   background: none;
   padding: 0.25em 0.5em;
   cursor: default;
-}
 
-.active {
-  cursor: pointer;
-  &:hover, &:focus {
-    opacity: 0.8;
-    color: var(--control-color, @control-color);;
-    outline-offset: -2px;
+  &.active {
+    cursor: pointer;
+    &:hover, &:focus {
+      opacity: 0.8;
+      color: var(--control-color, @control-color);
+      outline-offset: -2px;
+    }
   }
 }
 
-.copy-to-clipboard.copied,
-.clipboard-wrapper:hover .copy-to-clipboard.copied {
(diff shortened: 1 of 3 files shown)


nilmerg (changes requested):
This does not "resolve" the issue in Icinga DB Web. It refers to it at a maximum. Currently this looks like this: 
image

* Please only show the button upon hover of the parent (The rule for this must be available here!)
* Please fade "Copied" out
* The button's outline is placed weird, has different spacing left and right

asset/js/widget/CopyToClipboard.js
```diff
@@ -0,0 +1,67 @@
...
+                } else {
+                    copyText = el.innerText;
+                }
+            } else {
+                copyText = button.parentElement.innerText;
+            }
+
+            navigator.clipboard.writeText(copyText).then(() => {
+                let previousHtml = button.innerHTML
+                button.innerHTML = 'Copied';
```
nilmerg, line 48: Please fetch the label from a data attribute and translate it on the server.

asset/js/widget/CopyToClipboard.js
```diff
@@ -0,0 +1,67 @@
+define(["../notjQuery"], function ($) {
+
+    "use strict";
+
+    class CopyToClipboard {
+        constructor(button)
+        {
+            this.render(button);
+            $(button).on('click', null, this.onClick, this);
+
```
nilmerg, line 10: please remove that empty line

nilmerg (changes requested):

nilmerg (changes requested):
I just noticed in the linked PR that we litter the markup with IDs for no real reason. The thing with IDs is that they are (need..) to be unique, but this uniqueness is not required here.

We already have a guaranteed wrapper. So why not use it as point of reference?

In the button's `parentElement` (which is assumed to be always the wrapper) we lookup a direct child with the attribute `data-clipboard-source` and use it as source.

No fallback needed. If it's not there, an error should be logged. (We have no source that has no id, at this time)

Anyone using `attachTo` doesn't need to adjust the element then in any way, the data-attribute can be set automatically.

src/Widget/CopyToClipboard.php
```diff
@@ -0,0 +1,73 @@
...
+        $clipboardWrapper = new HtmlElement(
+            'div',
+            Attributes::create(['class' => 'clipboard-wrapper'])
+        );
+
+        if ($source->hasAttribute('id')) {
+            $button->addAttributes(['data-clipboard-source' => $source->getAttribute('id')->getValue()]);
+        } else {
+            $button->addAttributes(['data-clipboard-source' => 'parent']);
+        }
```
nilmerg, line 62: Remove this, the id of the source should be not of interest. Add the attribute to the source instead.

src/Widget/CopyToClipboard.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $clipboardWrapper->addHtml($source, $button);
+
+        return $clipboardWrapper;
+    }
+
+    public function assemble(): void
+    {
+        $this->setContent(new Icon('clone'));
```
nilmerg, line 71: setHtmlContent?

src/Widget/CopyToClipboard.php
```diff
@@ -0,0 +1,73 @@
...
+
+        if ($source->hasAttribute('id')) {
+            $button->addAttributes(['data-clipboard-source' => $source->getAttribute('id')->getValue()]);
+        } else {
+            $button->addAttributes(['data-clipboard-source' => 'parent']);
+        }
+
+        $clipboardWrapper->addHtml($source, $button);
+
+        return $clipboardWrapper;
```
nilmerg, line 66: Please set the wrapper really as wrapper and return void here.

nilmerg (changes requested):

asset/css/copy-to-clipboard.less
```diff
@@ -0,0 +1,35 @@
...
+.copy-to-clipboard {
+  appearance: none;
+  border: none;
+  opacity: 0;
+  background: none;
+  padding: 0.25em 0.5em;
+
+  &:hover, &:focus {
+    opacity: 0.8;
+    color: @control-color;
```
nilmerg, line 26:
```suggestion
    color: var(--control-color, @control-color);
```

nilmerg (changes requested):
With the plan in mind to introduce the behavior for this in Icinga Web, the feature won't be available unless the user upgrades to the new release. The click on the button then has no effect. So please let the constructor of `CopyToClipboard.js` add a class that enables the hover/focus effect. And add a `tabindex=-1` by default to the button, which the constructor should also remove. This prevents the button from being reachable by keyboard.

flourish86 (approved):
Fine from the design perspective.

asset/js/widget/CopyToClipboard.js
```diff
@@ -0,0 +1,55 @@
...
+                    button.innerText = button.dataset.copiedLabel;
+                    button.classList.add('persistent');
+
+                    // fade out after 3 second
+                    setTimeout(() => {
+                        // after 4 second, reset it.
+                        button.classList.add('copied');
+                        setTimeout(() => {
+                            button.classList.remove('copied');
+                            button.classList.remove('persistent');
```
nilmerg, line 37: Having two classes for this is somehow redundant. Use only one please.

nilmerg (changes requested):

nilmerg (approved):

PR #154, opened 2023-06-02
Add optional `$defaultSort` parameter for `SortControl::apply` method

raviks789:
This gives an option to set the default sort of sort control other than the `ipl\Orm\Model`s default sort as default.

### DIFF
--- a/src/Control/SortControl.php
+++ b/src/Control/SortControl.php
@@ -169,12 +169,13 @@ class SortControl extends Form
      * Sort the given query according to the request
      *
      * @param Query $query
+     * @param ?array|string $defaultSort
      *
      * @return $this
      */
-    public function apply(Query $query)
+    public function apply(Query $query, $defaultSort = null)
     {
-        $default = (array) $query->getModel()->getDefaultSort();
+        $default = $defaultSort ?? (array) $query->getModel()->getDefaultSort();
         if (! empty($default)) {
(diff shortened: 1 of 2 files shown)


src/Control/SortControl.php
```diff
@@ -169,12 +169,13 @@ public function getSort()
...
      * @param Query $query
+     * @param ?array|string $defaultSort
      *
      * @return $this
      */
-    public function apply(Query $query)
+    public function apply(Query $query, $defaultSort = null)
     {
-        $default = (array) $query->getModel()->getDefaultSort();
+        $default = $defaultSort ?? (array) $query->getModel()->getDefaultSort();
```
nilmerg, line 178: Previously `$default` was always an array, even if empty, due to the cast. Now if `$defaultSort` is not an array, `$default` isn't as well. I'm not sure the other code can handle this, have you checked?

raviks789: There is no change needed as `SortUtil::normalizeSortSpec` accepts both array or string as parameters and it is the only place where `$default` is used.

nilmerg (changes requested):
The `Controller` class here also has a `createSortControl` method, please adjust it as well.

nilmerg (changes requested):

nilmerg (approved):

PR #156, opened 2023-06-05
balls.less: Don't set the content color in `.ball-outline()`

sukhwinder33445:


nilmerg:
This applies to all icons with just an outline. i.e. ok, up and pending. @flourish86 Is it ok that it applies to pending states?

flourish86:
> This applies to all icons with just an outline. i.e. ok, up and pending. @flourish86 Is it ok that it applies to pending states?

Yes. 😊

flourish86 (approved):

nilmerg (changes requested):
Please change the Commit/PR description. It's not *unused*.

PR #157, opened 2023-06-07
Add icinga-icons src

flourish86:
As discussed, I added the src files for Icinga-icons


nilmerg (changes requested):
Icinga-Icons.svg|woff|ttf are already there. Please replace them at `asset/static/font`.

The CSS should also replace what's already inside `asset/css/icinga-icons.less`.

Are the demo files really required? Did you create them manually, or is this just another export by the icomoon app?

nilmerg (approved):

PR #159, opened 2023-06-20
Github Actions: Add Phpstan

sukhwinder33445:


yhabteab:
The remaining errors should be gone, once the phpstan branch of ipl-html is merged.

nilmerg (changes requested):
Not quite it seems

nilmerg (approved):

PR #166, opened 2023-07-10
state-badge.less: Add support for state `unreachable`

nilmerg:


asset/css/variables.less
```diff
@@ -48,7 +48,7 @@
 @state-down: @state-critical;
 @state-pending: #77aaff;
 @state-unknown: #aa44ff;
-@state-unreachable: @state-unknown;
+@state-unreachable: @gray-semilight;
```
nilmerg, line 51: That's a variable from Icinga Web, use one that's defined here (preferred) or define a new one (but then don't forget the light mode variable)

nilmerg:
Not required anymore.

PR #167, opened 2023-07-12
Introduce `Style` element

lippserd:
This PR introduces a new `Style` element to provide strict CSP compliant dynamic CSS when necessary.

**Note that it is still a work in progress, so it is not in the right namespace, is not documented, API is subject to change, and lacks unit tests.**

Example of creating style with some rulesets:

```
$style = new Style();
$style->add('.size', [
    'height' => '1em',
    'width'  => '1em'
]);
$style->add('.border', [
    'border' => '1px solid black'
]);
```

Output:

```
<style>
.size {
  height: 1em;
  width: 1em;
}
.border {
  border: 1px solid black;
}
</style>
```

Now we wrap the rules in a selector:

```
$style->setSelector('.parent');
```

Output:

```
<style>
.parent .size {
  height: 1em;
  width: 1em;
}
.parent .border {
  border: 1px solid black;
}
</style>
```

We can also use LESS variables and nested rulesets:

```
$style->setProperty('@black', 'black');
$style->addRuleset(
    LessRuleset::create('.widget', ['color' => '@black'])
        ->add('&.detail', ['font-size' => 'bigger'])
);
```

Output:

```
<style>
.parent .size {
  height: 1em;
  width: 1em;
}
.parent .border {
  border: 1px solid black;
}
.parent .widget {
  color: var(--black, #000000);
}
.parent .widget.detail {
  font-size: bigger;
}
</style>
```

Style can be automatically wrapped in a module scope:

```
$style->setModule(Icinga::app()->getRequest()->getModuleName());
```

Output:

```

<style>
.icinga-module.module-icingadb .parent .size {
  height: 1em;
  width: 1em;
}
.icinga-module.module-icingadb .parent .border {
  border: 1px solid black;
}
.icinga-module.module-icingadb .parent .widget {
  color: var(--black, #000000);
}
.icinga-module.module-icingadb .parent .widget.detail {
  font-size: bigger;
}
</style>
```

And for conformance with strict CSP, a nonce can be set:

```
$style->setNonce(Csp::getStyleNonce());
```

Output:

```
<style nonce="9MEcubtLrEm7isUUbl/l0A==">
...
</style>
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
     "ipl/orm": ">=0.5.2",
     "ipl/scheduler": ">=0.1.0",
     "ipl/stdlib": ">=0.12.0",
-    "fortawesome/font-awesome": "^6"
+    "fortawesome/font-awesome": "^6",
+    "wikimedia/less.php": "^3.2.1"
   },
   "require-dev": {
     "ipl/html": "dev-main",
--- a/src/LessRuleset.php
+++ b/src/LessRuleset.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace ipl\Web;
+
+use ArrayObject;
+use Less_Parser;
+
+class LessRuleset extends ArrayObject
+{
+    /** @var ?string */
+    protected $selector;
+
+    /** @var list<LessRuleset> */
+    protected $children = [];
+
+    public static function create(string $selector, array $properties): self
+    {
+        $ruleset = new static();
+        $ruleset->selector = $selector;
+        $ruleset->exchangeArray($properties);
+
+        return $ruleset;
+    }
+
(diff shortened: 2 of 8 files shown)


lippserd:
@nilmerg I'd like to discuss API and namespace.

nilmerg:
Thinking about using `lessc` directly, as suggested by @yhabteab. Inline style is mainly used for layout purposes and for this, theming and mode support isn't relevant. If we don't support that this way, devs are encouraged to define proper styles.

nilmerg (approved):
Since it doesn't depend on Icinga Web anymore, the namespace should be fine as it is now.

PR #194, opened 2023-08-31
CSS: Apply the same padding to filter input area as in icingadb

yhabteab:


sukhwinder33445 (approved):

nilmerg:
Requires a different solution: #201

PR #197, opened 2023-09-05
Add different redirect options

yhabteab:
### TODO
- [x] Currently, there is now way to determine whether the modal opener is `col1` or `col2` https://github.com/Icinga/icingaweb2/pull/5105
- [x] If the modal opener is `col1` it won't be refreshed when redirecting using `__CLOSE__`

Test commit:https://github.com/Icinga/icingaweb2-module-reporting/pull/194

### DIFF
--- a/src/Compat/CompatController.php
+++ b/src/Compat/CompatController.php
@@ -3,6 +3,7 @@
 namespace ipl\Web\Compat;
 
 use GuzzleHttp\Psr7\ServerRequest;
+use Icinga\Application\Version;
 use InvalidArgumentException;
 use Icinga\Web\Controller;
 use ipl\Html\BaseHtmlElement;
@@ -418,6 +419,70 @@ class CompatController extends Controller
         }
 
         $this->getResponse()->setHeader('X-Icinga-Extra-Updates', join(',', $extraUpdates));
+    }
+
+    /**
+     * Close the modal content and refresh the related view
+     *
+     * NOTE: If you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal,
+     * and refreshes the modal opener (regardless of whether it is col1 or col2).
+     *
+     * @param Url|string $url
+     * @param bool $refreshCol1 Whether to refresh col1 after the redirect. Is just for compatibility reasons and
+     *        won't be used with latest Icinga Web versions.
+     *
+     * @return never
+     */
+    public function closeModalAndRefreshRelatedView($url, bool $refreshCol1 = false)
+    {
+        if (version_compare(Version::VERSION, '2.12.0', '<')) {
+            if (! $url) {
+                throw new InvalidArgumentException('No redirect url provided');
+            }
+
+            if ($refreshCol1) {
+                $this->sendExtraUpdates(['#col1']);
+            }
+
+            $this->redirectNow($url);
+        } else {
+            $this->redirectNow('__REFRESH__');
+        }
+    }
+
+    /**
+     * Close the modal content and refresh all the remaining views
+     *
+     * NOTE: If you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal content. It refreshes
+     * the modal opener (expects to be always col2) and forces a refresh of col1.
+     *
+     * @param Url|string $url
+     *
+     * @return never
+     */
+    public function closeModalAndRefreshRemainingViews($url)
+    {
+        $this->sendExtraUpdates(['#col1']);
+
+        $this->closeModalAndRefreshRelatedView($url);
+    }
+
+    /**
+     * Redirect using `__CLOSE__`
+     *
+     * Change to a single column layout and refresh col1
+     *
+     * @return never
+     */
+    public function switchToSingleColumnLayout()
+    {
+        $this->redirectNow('__CLOSE__');
     }
 
     public function postDispatch()


yhabteab:
## TODO

- [ ] Sending extra updates for the `col2` container shouldn't open a new one if it isn't already there (has to be fixed in Icinga Web 2.).

src/Compat/CompatController.php
```diff
@@ -420,6 +421,52 @@ public function sendExtraUpdates(array $updates)
...
+            $this->redirectNow('__CLOSE__');
+        }
+    }
+
+    /**
+     * Redirect using `__CLOSE__`
+     *
+     * Change to layout 1 column and refresh col1
+     */
+    public function switchToLayout1Column(): void
```
nilmerg, line 465: *switchToSingleColumnLayout*

nilmerg (changes requested):
With https://github.com/Icinga/icingaweb2/pull/5108 it's now possible to refresh only the column that would also be affected by any other redirect such as `__CLOSE__`. (i.e. for modals, the opener)

Please rely on that for `closeCurrentAndRefreshRelatedView`. Though, rename it to `closeModalAndRefreshRelatedView`. I'd also suggest to add a second optional parameter which allows to define which column should be refreshed by an extra update in case of a legacy Web version.

Then, for the case where `#col1` and `#col2` should be refreshed, please add a new method. As name I'd suggest `closeModalAndRefreshRemainingViews` and assume that the modal is always shown in `#col2` so you can issue a `__REFRESH__` and an extra update for `#col1`. This is the only known case where refreshing both columns is required, so I think it's sufficient this way.

src/Compat/CompatController.php
```diff
@@ -420,6 +421,60 @@ public function sendExtraUpdates(array $updates)
...
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     *  This is supposed to be used in combination with a modal view and closes only the modal content,
+     * and refreshes the modal opener (regardless of whether it is col1 or col2).
+     *
+     * @param null|Url|string $url
+     * @param array<int|string, string> $extraUpdates See {@see static::sendExtraUpdates()} how to properly
+     *        provide extra updates
+     */
+    public function closeModalAndRefreshRelatedView($url = null, array $extraUpdates = []): void
```
nilmerg, line 437: aand here

src/Compat/CompatController.php
```diff
@@ -420,6 +421,60 @@ public function sendExtraUpdates(array $updates)
...
+    {
+        $this->sendExtraUpdates($extraUpdates);
```
nilmerg, line 439: these should only be sent if Web is < 2.12

src/Compat/CompatController.php
```diff
@@ -420,6 +421,60 @@ public function sendExtraUpdates(array $updates)
...
+     * Close the modal content and refresh the related view
+     *
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     *  This is supposed to be used in combination with a modal view and closes only the modal content,
+     * and refreshes the modal opener (regardless of whether it is col1 or col2).
+     *
+     * @param null|Url|string $url
+     * @param array<int|string, string> $extraUpdates See {@see static::sendExtraUpdates()} how to properly
```
nilmerg, line 434: and why should I provide them? Document the reason

src/Compat/CompatController.php
```diff
@@ -420,6 +421,60 @@ public function sendExtraUpdates(array $updates)
         $this->getResponse()->setHeader('X-Icinga-Extra-Updates', join(',', $extraUpdates));
     }
 
+    /**
+     * Close the modal content and refresh the related view
+     *
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     *  This is supposed to be used in combination with a modal view and closes only the modal content,
```
nilmerg, line 430: remove the leading space

src/Compat/CompatController.php
```diff
@@ -420,6 +421,60 @@ public function sendExtraUpdates(array $updates)
...
+    }
+
+    /**
+     * Close the modal content and refresh all the remaining views
+     *
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal content. It refreshes
+     * the modal opener (regardless of whether it is col1 or col2) and sends extra updates for col1.
```
nilmerg, line 459: nope, it expects the opener to be col2, otherwise the hardcoded update for col1 wouldn't make any sense

nilmerg (changes requested):

yhabteab:
GitHub actions fail with function never exit execution or throw exception.

nilmerg:
Then we'll add the same doc return type to ActionController::redirectNow()

yhabteab:
Requires https://github.com/Icinga/icingaweb2/pull/5110

src/Compat/CompatController.php
```diff
@@ -420,6 +421,69 @@ public function sendExtraUpdates(array $updates)
         $this->getResponse()->setHeader('X-Icinga-Extra-Updates', join(',', $extraUpdates));
     }
 
+    /**
+     * Close the modal content and refresh the related view
+     *
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
```
nilmerg, line 427: *If* you...

src/Compat/CompatController.php
```diff
@@ -420,6 +421,69 @@ public function sendExtraUpdates(array $updates)
...
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal content,
```
nilmerg, line 430: only the *modal,*

src/Compat/CompatController.php
```diff
@@ -420,6 +421,69 @@ public function sendExtraUpdates(array $updates)
...
+    }
+
+    /**
+     * Close the modal content and refresh all the remaining views
+     *
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal content. It refreshes
+     * the modal opener (expects to be always col2) and sends extra updates for col1.
```
nilmerg, line 462: and *forces a refresh of col1.*

src/Compat/CompatController.php
```diff
@@ -420,6 +421,69 @@ public function sendExtraUpdates(array $updates)
...
+     * NOTE: When you use this with older Icinga Web versions (< 2.12), you will need to specify a valid redirect url,
+     * that will produce the same result as using the `__REFRESH__` redirect with the latest Icinga Web version.
+     *
+     * This is supposed to be used in combination with a modal view and closes only the modal content,
+     * and refreshes the modal opener (regardless of whether it is col1 or col2).
+     *
+     * @param null|Url|string $url
+     * @param array<int|string, string> $extraUpdates Provide containers other than the redirect target to be refreshed
+     *        after the redirection. Is just for compatibility reasons and won't be used with latest Icinga Web
+     *        versions. See {@see static::sendExtraUpdates()} how to properly provide extra updates.
```
nilmerg, line 436: Thinking about this, I don't see the need for this being an array. A bool should suffice which allows the caller to define that also `col1` should be refreshed.

If the modal opener is col1, the url refreshes it. If the opener is col2, only col1 might be related. Especially since there's no col3 yet.

nilmerg (changes requested):
One last thing: The url shouldn't be optional, should it? Any module not passing it, only supports Icinga Web v2.12, which isn't released yet anyway. I don't think anyone developing a module will only support the latest major, and support for 2.11.x won't end soon enough.

nilmerg (approved):

PR #198, opened 2023-09-06
Link: Allow to optionally enable modal view

yhabteab:


nilmerg (changes requested):
I'd rather create a trait and use that in the `Link` class. The trait allows the user to optionally make the link a modal opener and since `Link` is the base for `ActionLink` and `ButtonLink` the same applies to them.

The reason for this is, that the javascript implementation for modals expect the opener to be an anchor with a `href` attribute. A plain `button` wouldn't work as `href` isn't a valid attribute for it.

So the only way to not have a *link* act as opener, but a button, is stylewise, and that's what `ButtonLink` is for.

The trait should have a public method `openInModal(): self` and applies the appropriate attributes there.

src/Widget/Link.php
```diff
@@ -80,4 +80,22 @@ public function createHrefAttribute()
...
+     * @return $this
+     */
+    public function openInModal(): self
+    {
+        $this->getAttributes()
+            ->set(
+                Attributes::create([
+                    'data-icinga-modal'   => true,
+                    'data-no-icinga-ajax' => true
+                ])
```
nilmerg, line 96: Why the detour with `Attributes::create`? Please set the attributes individually.

yhabteab: Well, why not? The `Attributes::set()` function accepts Attributes instance as well as array attributes, why do the attributes have to be set individually here?

nilmerg (approved):

PR #202, opened 2023-09-08
Introduce class `ipl\Web\Compat\StyleWithNonce`

nilmerg:


src/Compat/StyleWithNonce.php
```diff
@@ -0,0 +1,25 @@
...
+ * with Icinga Web &lt; 2.12 and with CSP support in &gt;= 2.12
+ */
+class StyleWithNonce extends Style
+{
+    public function getNonce(): ?string
+    {
+        if ($this->nonce === null) {
+            $this->nonce = version_compare(Version::VERSION, '2.12.0', '>=')
+                ? Csp::getStyleNonce()
+                : '';
```
yhabteab, line 20: Why empty string instead of null?

sukhwinder33445 (dismissed):

src/Compat/StyleWithNonce.php
nilmerg: Because otherwise the version compare and method call would needlessly occur in case of < 2.12

nilmerg: ..at every getter call

yhabteab (approved):

src/Compat/StyleWithNonce.php
```diff
@@ -0,0 +1,25 @@
...
+ * Use this class to define inline style which is compatible
+ * with Icinga Web &lt; 2.12 and with CSP support in &gt;= 2.12
+ */
+class StyleWithNonce extends Style
+{
+    public function getNonce(): ?string
+    {
+        if ($this->nonce === null) {
+            $this->nonce = version_compare(Version::VERSION, '2.12.0', '>=')
+                ? Csp::getStyleNonce()
```
nilmerg, line 19: oh well, should also return `?? ''` i guess

PR #208, opened 2023-09-18
Revert "Add missing css variable and state-ball for state `unreachable`"

nilmerg:
This reverts commit 15fd6b6985fa53bf42a7d57c13bbb3d889de55af.


nilmerg:
We don't need state badges with a state for unreachable anymore, as unreachable isn't visualized as state in Icinga DB Web anymore.

PR #210, opened 2023-09-29
CompatController: Add `getFilter` method

raviks789:
fixes Icinga/icinga-notifications-web#122


yhabteab (approved):

raviks789:
The `getFiler()` will be implemented in the notifications module just like in `icingadb-web`.

PR #217, opened 2024-03-20
BaseItemList: Emit events before and after adding the list item to the list

raviks789:
It would be helpful if lists could perform additional tasks before and after a list item is added to it. Hence the events `BEFORE_ITEM_ADD` and `ON_ITEM_ADD` would be desirable.

### DIFF
--- a/src/Common/BaseItemList.php
+++ b/src/Common/BaseItemList.php
@@ -15,6 +15,12 @@ abstract class BaseItemList extends BaseHtmlElement
 {
     use BaseFilter;
 
+    /** @var string Emitted while assembling the list after adding each list item */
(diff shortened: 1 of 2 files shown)


src/Common/BaseItemList.php
```diff
@@ -15,6 +15,11 @@ abstract class BaseItemList extends BaseHtmlElement
 {
     use BaseFilter;
 
+    /** @var string Emitted while assembling the list after adding each list item */
+    public const ON_ITEM_ADD = 'item-added';
+
+    public const BEFORE_ITEM_ADD = 'before-item-add';
```
nilmerg, line 21: Add documentation

nilmerg (changes requested):

PR #225, opened 2024-06-26
Move action-list from icingadb-web

sukhwinder33445:
Requires:

- https://github.com/Icinga/icingaweb2/pull/5238


sukhwinder33445:
Note to myself:

- Dashboard:  Click on an element in list 1, and before it can be loaded, click on an element in list 2 (list 1’s req is cancelled). You will see that list 2 displays the detailed view, but the element in list 1 is still marked as selected.
- Generalize the data attributes and Icinga dependencies.

PR #233, opened 2024-10-21
SearchControls: Add available `$preserveParams` as default to `(SearchBar/SearchEditor)suggestionUrl`

sukhwinder33445:
This way we no longer have to set the suggestionUrl manually, as in (Host/Service)groupController

### DIFF
--- a/src/Compat/SearchControls.php
+++ b/src/Compat/SearchControls.php
@@ -58,10 +58,13 @@ trait SearchControls
         $preserveParams = array_pop($params) ?? [];
         $redirectUrl = array_pop($params);
 
+        $requestUrlClone = $requestUrl->onlyWith($preserveParams);
+        $paramsToAdd = $requestUrlClone->getParams()->toArray(false);
+
         if ($redirectUrl !== null) {
-            $redirectUrl->addParams($requestUrl->onlyWith($preserveParams)->getParams()->toArray(false));
+            $redirectUrl->addParams($paramsToAdd);
         } else {
-            $redirectUrl = $requestUrl->onlyWith($preserveParams);
+            $redirectUrl = $requestUrlClone;
         }
 
         $filter = QueryString::fromString((string) $this->params)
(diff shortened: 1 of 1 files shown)


src/Compat/SearchControls.php
```diff
@@ -58,10 +58,13 @@ public function createSearchBar(Query $query, ...$params): SearchBar
         $preserveParams = array_pop($params) ?? [];
         $redirectUrl = array_pop($params);
 
+        $paramsToAdd = $requestUrl->onlyWith($preserveParams)->getParams()->toArray(false);
```
nilmerg, line 61: Only `addParams` requires an actual array. Don't transform parameters needlessly to an array.

src/Compat/SearchControls.php
```diff
@@ -83,7 +86,7 @@ public function createSearchBar(Query $query, ...$params): SearchBar
         if (method_exists($this, 'completeAction')) {
             $searchBar->setSuggestionUrl(Url::fromPath(
                 "$moduleName/$controllerName/complete",
-                ['_disableLayout' => true, 'showCompact' => true]
+                array_merge($paramsToAdd, ['_disableLayout' => true, 'showCompact' => true])
```
nilmerg, line 89: Since you will always have a UrlParams object after not transforming them always to an array, you can just call `$paramsToAdd->set(…)` twice.

src/Compat/SearchControls.php
```diff
@@ -58,10 +58,13 @@ public function createSearchBar(Query $query, ...$params): SearchBar
...
         $redirectUrl = array_pop($params);
 
+        $paramsToAdd = $requestUrl->onlyWith($preserveParams)->getParams()->toArray(false);
+
         if ($redirectUrl !== null) {
-            $redirectUrl->addParams($requestUrl->onlyWith($preserveParams)->getParams()->toArray(false));
+            $redirectUrl->addParams($paramsToAdd);
         } else {
-            $redirectUrl = $requestUrl->onlyWith($preserveParams);
+            $redirectUrl = clone $requestUrl;
```
nilmerg, line 66: You're cloning again, for no reason, here. Clone only once.

nilmerg (changes requested):
My comments apply to both methods.

nilmerg (approved):

PR #234, opened 2024-10-24
BaseItemList: Add setter/getter to set empty state message

raviks789:
To make the empty state message customizable for lists, we need setter/getter methods in `BaseItemList`

### DIFF
--- a/src/Common/BaseItemList.php
+++ b/src/Common/BaseItemList.php
@@ -33,6 +33,9 @@ abstract class BaseItemList extends BaseHtmlElement
 
     protected $tag = 'ul';
 
+    /** @var ?string Message to show if the list is empty */
+    protected $emptyStateMessage;
+
     /**
      * Create a new item  list
      *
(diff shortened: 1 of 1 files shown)


src/Common/BaseItemList.php
```diff
@@ -33,6 +33,9 @@ abstract class BaseItemList extends BaseHtmlElement
 
     protected $tag = 'ul';
 
+    /** @var ?string Message when the list is empty */
```
nilmerg, line 36: *Message to show if…*

src/Common/BaseItemList.php
```diff
@@ -76,6 +79,34 @@ protected function createListItem(object $data)
         return new $className($data, $this);
     }
 
+    /**
+     * Get message to show for empty list
```
nilmerg, line 83: *…if the list is empty*

src/Common/BaseItemList.php
```diff
@@ -76,6 +79,34 @@ protected function createListItem(object $data)
...
+    {
+        if ($this->emptyStateMessage === null) {
+            return t('No items found.');
+        }
+
+        return $this->emptyStateMessage;
+    }
+
+    /**
+     * Set message for empty list
```
nilmerg, line 97: *…to show if the list is empty*

nilmerg (changes requested):

nilmerg (approved):

PR #236, opened 2024-11-21
SearchControls: Don't use UrlParams instance for setEditorUrl()

sukhwinder33445:


src/Compat/SearchControls.php
```diff
@@ -93,7 +93,7 @@ public function createSearchBar(Query $query, ...$params): SearchBar
         if (method_exists($this, 'searchEditorAction')) {
             $searchBar->setEditorUrl(Url::fromPath(
                 "$moduleName/$controllerName/search-editor"
-            )->setParams($redirectUrl->getParams()));
+            )->setParams($redirectUrl->getParams()->toArray(false)));
```
nilmerg, line 96: Does clone not work?

nilmerg (approved):
Note that I do not approve this to justify what you did [here](https://github.com/Icinga/icingadb-web/blob/dependencies/application/controllers/RedundancygroupController.php#L187-L189).

The reason is rather, that the redirect URL can be provided by the caller. In this case the caller is responsible for the parameters, in which case it's bad practice to pass them on to another URL or even adjust them without cloning.

nilmerg:
@cla-bot check

PR #240, opened 2025-01-16
item-list.less: Always apply `calculated &nbsp;` margin to `.title` content

sukhwinder33445:


nilmerg:
The margin you're adjusting here is where it is for a reason. It is tightly connected with the `.title` being a flex container. As such, white space inside it collapses and the margin restores it.

If the margin is always (i.e. also on non-flex containers) applied, it is combined with white space and the resulting gaps are too wide.

The margin rules however, as they currently exist, are applicable to only a very specific structure: `<html> text <html> text`

Any `<html> <html>` structure will not work with it. Is this the reason why you wanted to change it?

nilmerg:
I saw that comment already, but it doesn't explain why it's required in detailed mode.

sukhwinder33445:
As disscussed offline, this change is not required.

PR #245, opened 2025-01-20
item-list.less: Don't unset `margin-right` for `:only-child`

sukhwinder33445:
RedundancyGroupListItem's title contain only one element. In this his case, the margin-right should not be unset.

ref https://github.com/Icinga/icingadb-web/pull/1098

### DIFF
--- a/asset/css/list/item-list.less
+++ b/asset/css/list/item-list.less
@@ -69,7 +69,7 @@
         margin-left: 0;
       }
 
-      &:last-child {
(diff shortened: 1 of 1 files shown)


sukhwinder33445:
Discussed with Johannes, this is a redundancy list specific thing, so only needs to be introduced for redundancy lists. I will add it with https://github.com/Icinga/icingadb-web/pull/1098.

Issue #243, opened 2025-01-20
Increase obviousness of how to remove a term in a term-input


nilmerg:
At the moment a term can be removed by clicking on it. This must not change. However, the indication of this functionality needs to be improved. Hovering over a term shows a trash icon on the left hand. The color doesn't change.

The trash icon must be shown in the middle over the term and have a reddish color. The term's label must have a grayish color then.

flourish86:
What do you think of these solutions?

I see some advantages over what you've suggested:

* having a dedicated click area for deletion prevents the user from accidentally removing items
* A label makes it clear what will happen
* The user icon is part of the label and is independent from the delete action

I'm aware of the larger implementation effort, since there are more changes to do. But since we're working on "global" element, I think it's worth it.

1) Compact: Hovering makes delete button appear

https://github.com/user-attachments/assets/26c56c70-0957-4174-8999-c118aa3f72c4

2) Always visible delete button. In case of Contacts in Notifications we should have enough space

https://github.com/user-attachments/assets/dd2d7e0b-e986-46c0-a818-a3168c7fb62e


If you insist on keeping the whole element clickable for deletion, I'd suggest this design, though.

3) Hover button and label replace label

https://github.com/user-attachments/assets/d54b3400-0658-49aa-a822-587f4e5622bb

PR #249, opened 2025-02-19
Display delete action when hovering over read-only term-input entries

ncosta-ic:
This PR changes the way by which the user gets hinted about the possibility to remove an entry out of a read-only list of term-input values.

Resolves: https://github.com/Icinga/ipl-web/issues/243

### DIFF
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -332,7 +332,7 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
         top: 0;
         line-height: normal;
         padding: 0.5em;
-        cursor: pointer;
+        pointer-events: none;
 
         .remove-action-label {
           .text-ellipsis();
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -75,6 +75,15 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   }
 }
 
+.remove-action {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background: @search-term-remove-action-bg;
+  color: @search-term-remove-action-color;
+  .rounded-corners(0.25em);
+}
+
 .search-suggestions {
   background: var(--suggestions-bg, @suggestions-bg);
   color: var(--suggestions-color, @suggestions-color);
@@ -167,7 +176,7 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   label {
     position: relative;
     display: inline-block;
-    min-width: 2em;
+    min-width: 6em;
     height: 100%;
 
     &::after,
@@ -210,6 +219,11 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
     label {
       margin-right: 1px;
     }
+  }
+
+  &.read-only [data-index] .remove-action {
+    line-height: 20/12;
+    padding: var(--term-padding-v) var(--term-padding-h);
   }
 }
 
@@ -281,7 +295,6 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
       position: relative;
 
       input {
-        padding-left: 1.5em;
         text-align: center;
         cursor: pointer;
 
@@ -307,6 +320,24 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
         z-index: 1;
         top: 85%;
         left: .5em;
+      }
+
+      .remove-action {
+        visibility: visible;
+        position: absolute;
+        width: 100%;
+        top: 0;
+        line-height: normal;
+        padding: 0.5em;
+        cursor: pointer;
+
+        .remove-action-label {
+          .text-ellipsis();
+        }
+      }
+
+      &:not(:hover) .remove-action {
+        visibility: hidden;
       }
     }
   }
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -58,30 +58,33 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   cursor: not-allowed;
 }
 
-.invalid-reason {
-  padding: .25em;
-  .rounded-corners(.25em);
-  border: 1px solid black;
-  font-weight: bold;
-  background: var(--search-term-invalid-reason-bg, @search-term-invalid-reason-bg);
+.term-input-area {
+  .invalid-reason {
+    padding: .25em;
+    .rounded-corners(.25em);
+    border: 1px solid black;
+    font-weight: bold;
+    background: var(--search-term-invalid-reason-bg, @search-term-invalid-reason-bg);
 
-  opacity: 0;
-  visibility: hidden;
-  transition: opacity 2s, visibility 2s;
-  &.visible {
-    opacity: 1;
-    visibility: visible;
-    transition: none;
+    opacity: 0;
+    visibility: hidden;
+    transition: opacity 2s, visibility 2s;
+
+    &.visible {
+      opacity: 1;
+      visibility: visible;
+      transition: none;
+    }
   }
-}
 
-.remove-action {
-  display: flex;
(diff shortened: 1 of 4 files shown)


asset/css/search-base.less
```diff
@@ -305,6 +323,29 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
...
+
+      .delete-action {
+        visibility: visible;
+        display: inline-flex;
+        position: absolute;
+        z-index: 1;
+        width: 100%;
+        top: 0;
+        left: 0;
+        height: var(--delete-action-height, 100%);
```
sukhwinder33445, line 335: Please add `cursor: pointer;`

sukhwinder33445 (changes requested):

asset/css/variables.less
```diff
@@ -66,6 +66,8 @@
 @search-term-highlighted-bg: @base-primary-bg;
 @search-term-highlighted-color: @default-text-color-inverted;
 @search-term-drag-border-color: @base-gray;
+@search-term-delete-action-bg: @state-critical;
```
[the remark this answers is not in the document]
ncosta-ic: I don't really see the benefit here, as long as there's no multiple references to it. Same could be said about line 61 and 62 :thinking:

src/FormElement/TermInput/TermContainer.php
```diff
@@ -57,9 +62,30 @@ protected function assemble()
...
+                                    Attributes::create([
+                                        'class' => 'delete-action-label'
+                                    ]),
+                                    new Text($this->getAttribute('delete-action-label')->getValue())
+                                )
+                            ])
+                        )
+                    )
                 );
+                $label->addHtml(new HtmlElement('span', Attributes::create(['class' => 'invalid-reason'])));
```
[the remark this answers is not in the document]
ncosta-ic: This would add it as a children of `remove-action`, which I don't want.

asset/css/search-base.less
```diff
@@ -264,6 +276,10 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
       }
     }
   }
+
+  &.read-only [data-index] > .remove-action {
```
sukhwinder33445, line 280: Please do not set a higher specificity here. It should be the same as the rule for non-vertical. So please remove `>`.

asset/css/search-base.less
```diff
@@ -277,8 +293,11 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
     [data-index] {
       position: relative;
 
+      &:not(:hover) > .remove-action {
```
sukhwinder33445, line 296: Please move it a few lines down, after or before the definition of `.remove-action` for better readability.

asset/css/search-base.less
```diff
@@ -75,6 +75,18 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   }
 }
 
+.remove-action {
+  display: flex;
+  text-align: center;
```
sukhwinder33445, line 80: `text-align` seems to be superfluous.

asset/css/search-base.less
```diff
@@ -305,6 +324,21 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
...
         left: .5em;
       }
+
+      .remove-action {
+        visibility: visible;
+        position: absolute;
+        z-index: 1;
+        width: 100%;
+        top: 0;
+        left: 0;
```
sukhwinder33445, line 334: You can move all these properties, which are always required (vertical or not), to the above definition for better readability.

sukhwinder33445 (changes requested):

asset/css/search-base.less
sukhwinder33445: Or even better, the above definition can be merged into this one, as the ‘Remove’ button is only needed if the input field is `read-only`.

src/FormElement/TermInput/TermContainer.php
```diff
@@ -33,6 +38,11 @@ public function __construct(TermInput $input)
 
     protected function assemble()
     {
+        if ($this->input->getReadOnly()) {
```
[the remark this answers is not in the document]
ncosta-ic: I'm not going to set that attribute for each loop iteration...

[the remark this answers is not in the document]
sukhwinder33445: Ohh, somehow I overlooked the loop scope. You're right.
Optional: I would still omit the call `getAttribute()->getValue()` for the label.

asset/css/search-base.less
ncosta-ic: Which definitions are you referring to with "above definition"? The one starting on line 78? I placed it there because it's considered to be style relevant and not layout relevant.

src/FormElement/TermInput/TermContainer.php
```diff
@@ -57,9 +62,30 @@ protected function assemble()
...
+                        'div',
+                        Attributes::create([
+                            'class' => 'delete-action'
+                        ]),
+                        new HtmlElement(
+                            'div',
+                            Attributes::create([
+                                'class' => 'delete-action-content'
+                            ]),
+                            ...Html::wantHtmlList([
```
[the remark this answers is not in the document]
ncosta-ic: I didn't. You were commenting on an old git state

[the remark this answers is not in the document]
sukhwinder33445: No 🤔 , i commented on the latest. I meant that the call to `...Html::wantHtmlList()` is not necessary here, you can remove it.

asset/css/search-base.less
```diff
@@ -75,6 +75,18 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
...
+.remove-action {
+  display: flex;
+  text-align: center;
+  justify-content: center;
+  align-items: center;
+  background: var(--search-term-remove-action-bg, @search-term-remove-action-bg);
+  color: var(--search-term-remove-action-color, @search-term-remove-action-color);
+  border: none;
+  border-radius: 0.25em;
+  padding: 0.25em 0.5em 0.25em 0.5em;
```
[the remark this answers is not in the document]
ncosta-ic: We agreed on removing the padding, as it's only set to mirror the input element, but since we're using a vertical and horizontal alignment, it never actually changes the rendering.

sukhwinder33445 (dismissed):
LGTM

asset/css/search-base.less
```diff
@@ -305,6 +317,23 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
...
         left: .5em;
       }
+
+      .remove-action {
+        visibility: visible;
+        position: absolute;
+        width: 100%;
+        top: 0;
+        height: ~"calc(2em + 1.75px)"; // respect the additional scrollbar in horizontal (default) mode
+        cursor: pointer;
```
sukhwinder33445, line 327: Please rebase, and remove the height.
Add the following instead:

```css
        line-height: 20/12; // same as input
        padding: var(--term-padding-v) 0; // same as input top/bottom padding
```

asset/css/search-base.less
```diff
@@ -308,6 +321,24 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
...
       }
+
+      .remove-action {
+        visibility: visible;
+        position: absolute;
+        width: 100%;
+        top: 0;
+        line-height: normal;
+        padding: 0.5em;
+        cursor: pointer;
```
nilmerg, line 332: It currently hides the validation message in case the value is invalid. This container should not catch focus hence should use `pointer-events: none` instead of this.

nilmerg (changes requested):

nilmerg (approved):

nilmerg (approved):

PR #251, opened 2025-03-07
SearchBar: Fix scrollbar position in default-layout controls

nilmerg:
The `overflow-x: overlay` declaration is obsolete, since no browser (not even chrome) supports it anymore.


sukhwinder33445 (approved):
👍🏽

PR #257, opened 2025-03-21
TermInput: Keep original minimum width of terms

### LINKED ISSUE #256: Searchbar term width too large
Since #249 short searchbar terms look as follows:

Image

nilmerg:
The remove label is now a title which allows terms to be as narrow as possible again.

justify-content which was previously used to center the label's contents cannot be used anymore though. It has the effect of pushing overflowing content
also to the left, which causes the icon to not be
centered at the minimum width. Plain old, but working, auto margins are the solution here.

fixes #256

### DIFF
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -78,9 +78,6 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   }
 
   .remove-action {
-    display: flex;
-    justify-content: center;
-    align-items: center;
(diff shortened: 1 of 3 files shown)


ncosta-ic (approved):
Your idea is way more elegant than what I've proposed in my PR...
image
 
This looks really good now as it still shows the icon correctly on a single-char input.
Double-checked it for Gecko (Firefox), Blink (Chrome) and WebKit (Safari / GNOME Web) and it looks and works fine.

PR #260, opened 2025-04-01
Fix that the sort control ignores a default

### LINKED ISSUE #242: SortControl is not working properly
## Bug
If the optional `$defaultSort` param is provided to the `Controller::createSortControl()`, the query is sorted accordingly, but the dropdown still shows the wrong option as selected (`$columns` first value).

## Expected Behavior
The correct value is displayed in the drop-down list.

## Further Suggestions
 - Throw error if the given `$defaultSort` is not present in the `$columns`.

nilmerg:
fixes #242

### DIFF
--- a/src/Control/SortControl.php
+++ b/src/Control/SortControl.php
@@ -222,7 +222,16 @@ class SortControl extends Form
         return $this;
     }
 
-    protected function assemble()
+    /**
+     * Prepare the visual representation of the sort control
+     *
+     * This is called just before rendering happens. What is being done here, doesn't influence validity in any way.
+     * So there is no need to have the result already at hand during validation. Instead, delaying it allows
+     * to influence the visual result as long as possible.
+     *
+     * @return void
+     */
+    protected function prepareContent(): void
     {
         $columns = $this->getColumns();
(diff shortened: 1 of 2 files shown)


sukhwinder33445 (commented):
The `setDefault()` is ignored when called by the controller, for example:
```php
$controller->createSortControl(...)->setDefault('value'); // SortControl still has no default OR the wrong one from model
```

SortControlTest uses an deprecated method, the following message is displayed when running the test:
`assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.`

nilmerg:
> The setDefault() is ignored when called by the controller

It's not ignored. It's pre selected for me. It isn't applied on the query, of course, as for that it's set way too late. For that to influence one has to use the second parameter for `createSortControl`.

> SortControlTest uses an deprecated method

phpunit 10 requires php 8.1+, unless we drop support for everything below that, this message won't disappear.

sukhwinder33445 (dismissed):

sukhwinder33445 (approved):

PR #264, opened 2025-04-04
Fix that search controls do not preserve url filters

### LINKED ISSUE #263: Search controls do not preserve url filters
An `ipl\Web\Url` can possess a filter. Search controls however, i.e the search bar, search editor and continue with control, accept such but override it with their own.

nilmerg:
fixes #263


raviks789 (approved):
LGTM!

PR #269, opened 2025-04-22
Provide a way to go back to the first page in case the current page is empty

### LINKED ISSUE #267: Provide a way to go back to the first page in case the current page is empty
The pagination control should provide a way to go back to the last page, in case the current page is empty. This is normally not possible to achieve using the UI, as it only suggests valid pages.

Though, by following a link a user got from someone, the resulting entries might be fewer than previously and hence the list will also be empty.

In case there are multiple pages available, the user then has the option to choose another page. However, if there is only a single page, the control doesn't provide any option to switch to it.

A simple link allowing to clear the page parameter should be shown.

Al2Klimov:
By following a link a user got from someone, the resulting entries might be fewer than previously and hence the list will also be empty.

If there is only a single page, the control didn't provide any option to switch to it until now.

Now a simple link allows to clear the page parameter.

fixes #267

### DIFF
--- a/src/Control/PaginationControl.php
+++ b/src/Control/PaginationControl.php
@@ -470,12 +470,14 @@ class PaginationControl extends BaseHtmlElement
             'title' => t('Go to page …')
         ]);
 
-        if ($currentPageNumber === 1 || $currentPageNumber === $this->getPageCount()) {
+        $pageCount = $this->getPageCount();
+
+        if ($currentPageNumber === 1 || $currentPageNumber === $pageCount) {
             $select->add(Html::tag('option', ['disabled' => '', 'selected' => ''], '…'));
         }
 
-        if (2 <= $this->getPageCount() - 1) {
-            foreach (range(2, $this->getPageCount() - 1) as $page) {
+        if (2 <= $pageCount - 1) {
+            foreach (range(2, $pageCount - 1) as $page) {
                 $option = Html::tag('option', [
                     'value' => $page
                 ], $page);
@@ -488,7 +490,7 @@ class PaginationControl extends BaseHtmlElement
             }
         }
 
-        if ($currentPageNumber > $this->getPageCount()) {
+        if ($currentPageNumber > $pageCount) {
             $select->add(Html::tag('option', [
                 'value' => $currentPageNumber,
                 'selected' => '',
--- a/src/Control/PaginationControl.php
+++ b/src/Control/PaginationControl.php
@@ -474,16 +474,18 @@ class PaginationControl extends BaseHtmlElement
             $select->add(Html::tag('option', ['disabled' => '', 'selected' => ''], '…'));
         }
 
-        foreach (range(2, $this->getPageCount() - 1) as $page) {
-            $option = Html::tag('option', [
-                'value' => $page
-            ], $page);
+        if (2 <= $this->getPageCount() - 1) {
+            foreach (range(2, $this->getPageCount() - 1) as $page) {
+                $option = Html::tag('option', [
+                    'value' => $page
+                ], $page);
 
-            if ($page == $currentPageNumber) {
-                $option->addAttributes(['selected' => '']);
+                if ($page == $currentPageNumber) {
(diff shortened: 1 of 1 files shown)


src/Control/PaginationControl.php
```diff
@@ -496,6 +496,17 @@ protected function createPageSelectorItem()
     protected function assemble()
     {
         if ($this->getPageCount() < 2) {
+            if ($this->getCurrentPageNumber() > 1) {
+                $this->add(Html::tag(
```
sukhwinder33445, line 500: I would suggest to use `addHtml()` instead of `add()`.
Please use the `ipl\Web\Widget\Link` class instead of creating `Html::tag('a'...)`

sukhwinder33445 (commented):
Screenshot 2025-04-29 at 10 06 11

If only the link is displayed, it looks strange. I think we should still show a disabled pagination control.

Any suggestions @nilmerg @flourish86 ?

Al2Klimov:
Why a _disabled_ pagination control? Why not such:

<< 1 42 >>

Where e.g 42 is your current (empty) page and 1 allows you to switch to a (hopefully) non-empty page.

src/Control/PaginationControl.php
```diff
@@ -496,6 +497,14 @@ protected function createPageSelectorItem()
     protected function assemble()
     {
         if ($this->getPageCount() < 2) {
+            if ($this->getCurrentPageNumber() > 1) {
+                $this->addHtml(new Link(
```
sukhwinder33445, line 501: Since the pagination numbers are blue (represent action links), I would suggest doing the same here. To do this, please use the class `ActionLink` instead of `Link`.

```php
$this->addHtml(new ActionLink(
    $this->translate('Invalid page, click here to reset'),
    $this->url->without('page'),
    null, //  you can even specify a suitable symbol
    ['title' => $this->createLabel(1)]
));
```

flourish86:
> Any suggestions @nilmerg @flourish86 ?

I could figure out the following scenarios for this issue:
1)  There is no pagination rendered if all items (for the currently set limit, with no explicit limit set defaulting to 25) fit on one page (no page param set, defaulting to page=1)
=> no problem here
Screenshot 2025-04-29 at 11 38 59

2) There is an empty state rendered and no pagination rendered if you navigate to an out of range page
=> It's not ovbious for users, what's wrong 
=> There's no obvious way to get back to useful results
Screenshot 2025-04-29 at 11 40 53

Since there probably aren't any issues for case 1) I see different solutions for case 2)
a) "Redirect" to the last available page automatically
b) notice the user, that the page is not available and offer a navigation item that takes him to results, displaying the pagination control as a known element would be helpful

this could look like this
Frame 1

flourish86:
lgtm.

I wonder, if we should mark the wrong page option somehow. To add an "(!)" to the option text or highlight the select element would make it more obvious that something is wrong. But I think for now that's enough.

flourish86 (dismissed):

Al2Klimov:
> I think for now that's enough.

Absolutely!

> we should mark the wrong page option somehow.

It's a selected and disabled \<option>. Probably not the only one in the IW2 universe, so we could highlight such in general. E.g with red font color. But you're our CSS master, so you know better _how_ to style it.

flourish86:
I discussed the topic offline with @nilmerg. The current solution is fine, with one exception.

Can you please hide the pagination control for the following case:
* user navigated to a page, that is out of range
* there's only one page ?

Test Link (env with <5000 hosts)
`…/icingaweb2/icingadb/hosts?limit=5000&page=35`

This is the current state from this PR:
Screen Shot 2025-05-06 at 09 49 51

and should be changed to this
Screen Shot 2025-05-06 at 09 50 52

Al2Klimov:
Because the paginator disappears once back on page one? Then I'd **always** show the paginator. Would have the positive effect that the user doesn't search for the paginator, knowing there's only one page. :)

nilmerg (approved):

PR #276, opened 2025-05-06
`PaginationControl`: Add getter `getEmptyStateMessage()` to provide pagination reset link

sukhwinder33445:


src/Widget/ItemList.php
```diff
@@ -31,7 +33,7 @@ class ItemList extends BaseHtmlElement
     /** @var string */
     private $itemLayoutClass = ItemLayout::class;
 
-    /** @var ?string Message to show if the list is empty */
+    /** @var ?ValidHtml Message to show if the list is empty */
     protected $emptyStateMessage;
```
sukhwinder33445, line 37: I'm thinking about renaming it to `$emptyStateContent` including the getters/setters.  What is your opinion @nilmerg ?

nilmerg: I'd rather not change the method name. Thus neither the property name.

src/Widget/ItemList.php
```diff
@@ -106,27 +108,27 @@ public function getItemLayout($item): ItemLayout
...
-     * Set message to show if the list is empty
+     * Set message content to show if the list is empty
      *
-     * @param string $message
+     * @param ?ValidHtml $message
      *
      * @return $this
      */
-    public function setEmptyStateMessage(string $message): self
+    public function setEmptyStateMessage(?ValidHtml $message): self
```
nilmerg, line 131: Use the `mixed` type (yeah, it's php 8 only but we raise the min requirement soon anyway) and use `Html::wantHtml` inside.

src/Control/PaginationControl.php
```diff
@@ -122,6 +126,27 @@ public function setPageParam($pageParam)
...
+    {
+        $currentPage = $this->getCurrentPageNumber();
+        $pageCount = $this->getPageCount();
+        if ($pageCount && $pageCount < $currentPage) {
+            return (new HtmlDocument())
+                ->addHtml(new Text(sprintf($this->translate('Page %d is out of range'), $currentPage)))
+                ->addHtml(new ButtonLink(
+                    $this->translate('Navigate to first page'),
+                    $this->url->remove(['page'])
+                ));
```
nilmerg, line 144: Please use a `TemplateString` here

nilmerg (changes requested):

nilmerg (changes requested):

nilmerg (changes requested):
This currently has two problems:

1. The change introduced by #269 has no effect anymore
2. Clicking on the "Navigate to first page" navigates to the last page

nilmerg (approved):

PR #281, opened 2025-05-14
Set line-clamp to 5

### LINKED ISSUE #279: Detailed item layout mismatching height and line-clamp
The calculated height of the `detailed-item-layout` does not match the number of lines passed to `.line-clamp()`. The `line-height` is `1.5em`.

```less
max-height: 7.5em; /* 5 lines */
.line-clamp(4);
```

The actual output is **four** lines long.

jrauh01:
And remove the superfluous max-height because line-clamp already limits the height.

Fixes #279

### DIFF
--- a/asset/css/item-layout.less
+++ b/asset/css/item-layout.less
@@ -86,12 +86,10 @@
   }
   &.detailed-item-layout > .main > .caption {
     display: block;
-    height: auto;
     overflow: hidden;
-    max-height: 7.5em; /* 5 lines */
     position: relative;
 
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
There may be browser which don't support line clamping ;)

jrauh01:
I tested it on Chrome, Firefox and Safari. All of them worked fine. (https://caniuse.com/?search=line-clamp)

nilmerg (approved):

PR #282, opened 2025-05-14
Limit default item layout caption to one line

jrauh01:
Reduce the caption from **two** lines to **one** line in the default item layout.


jrauh01:
Screenshot 2025-05-14 at 15 27 39

PR #284, opened 2025-05-15
ButtonLink: Allow to disable the button

sukhwinder33445:


src/Widget/ButtonLink.php
```diff
@@ -11,4 +11,43 @@ class ButtonLink extends ActionLink
...
+
+    /**
+     * Get whether the button is disabled
+     *
+     * @return bool
+     */
+    public function isDisabled(): bool
+    {
+        return $this->isDisabled;
+    }
```
nilmerg, line 40: I don't like the idea of having a button (link) that is disabled but doesn't have any explanation for this. Since the linked Notifications Web pull request adds a `title` attribute on its own, if it disables the button, I'd suggest to introduce another method: `ButtonLink::disable(string $explanation): static`

asset/css/action-link-and-button-link.less
```diff
@@ -0,0 +1,25 @@
...
+    background: @default-action-button-hover-bg;
+    text-decoration: none;
+  }
+
+  &[disabled] {
+    background: @gray-light;
+    color: @disabled-gray;
+    cursor: not-allowed;
+
+    .user-select(none);
```
nilmerg, line 23: Unreferenced mixin.

nilmerg (changes requested):
```css
// Icinga Web < 2.13

.button-link[aria-disabled="true"]:hover {
  background: none;
}
```

This in compat.less, avoids issues with older Web releases and

nilmerg:
And rebase please. There are new checks in place.

nilmerg (approved):
Looks fine now. Though, in case the button link is opened in a modal, we need https://github.com/Icinga/icingaweb2/pull/5398, otherwise the cursor/title is invisible once clicked and we have an error in the console.

PR #285, opened 2025-05-15
Clicking submit button should submit the form with term input

raviks789:
Previously, the first click of submit button in a form with with the term input was triggering an auto submit, instead of submitting the form. This fix resolves it.

**Before**:

https://github.com/user-attachments/assets/de4964ef-c7cb-47f1-8986-e17db8066f07

**After**:

https://github.com/user-attachments/assets/64bce31f-231e-43d2-802d-9b571713144c

### DIFF
--- a/asset/js/widget/TermInput.js
+++ b/asset/js/widget/TermInput.js
@@ -319,7 +319,12 @@ define(["../notjQuery", "../vendor/Sortable", "BaseInput"], function ($, Sortabl
         }
 
         onButtonClick(event) {
-            if (! this.hasSyntaxError()) {
+            // Exchange terms only when an Enter key is pressed while not in the term input.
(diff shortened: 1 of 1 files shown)


asset/js/widget/TermInput.js
```diff
@@ -319,7 +319,12 @@ define(["../notjQuery", "../vendor/Sortable", "BaseInput"], function ($, Sortabl
         }
 
         onButtonClick(event) {
-            if (! this.hasSyntaxError()) {
+            // Exchange terms only when an Enter key is pressed while in term input.
```
nilmerg: *while **not** in the term input*

At least that is the purpose of this event handler.

nilmerg (changes requested):

nilmerg (approved):

PR #286, opened 2025-05-15
Revert "Remove redundant selectors"

nilmerg:
This reverts commit 64c8d6b4265665771a38182aeb94bb9267d5410a.

Host list Icinga DB Web:
image
image

Specificity is sometimes a bitch.


jrauh01 (changes requested):
Looks like we indeed need the more specific selector. Maybe add a comment with an explanation why we need it?

PR #293, opened 2025-06-11
Suggestions: Do not rely on content to detect exact matches

### LINKED ISSUE #265: SearchBar: Suggestions error, for search terms in range 10 to 19
https://github.com/user-attachments/assets/0006bdf0-d6de-4fc7-a7ed-dcfd1833fd41

nilmerg:
Extending classes might adjust content beforehand and then the first `li` won't necessarily include a `button`.

fixes #265


raviks789 (commented):
You could also remove unused `ipl\Html\FormElement\InputElement`.

raviks789 (approved):
LGTM!

PR #296, opened 2025-06-12
variables.less: Add scrollbar-color variable

sukhwinder33445:


sukhwinder33445:
No longer required, see https://github.com/Icinga/icingaweb2/pull/5379#issuecomment-4125005774

PR #299, opened 2025-07-02
CsrfCounterMeasure: accept Sec-Fetch-Site header, if available, instead of token

Al2Klimov:
This is especially useful if the session and token change suddenly, e.g. due to mod_auth_openidc.

refs https://github.com/Icinga/icingaweb2/issues/5224

### DIFF
--- a/src/Common/CsrfCounterMeasure.php
+++ b/src/Common/CsrfCounterMeasure.php
@@ -26,6 +26,12 @@ trait CsrfCounterMeasure
             'ignore'        => true,
             'required'      => true,
             'validators'    => ['Callback' => function ($token) use ($uniqueId, $hashAlgo) {
+                switch ($_SERVER['HTTP_SEC_FETCH_SITE'] ?? '') {
+                    case 'same-origin': // same scheme, host and port
+                    case 'none': // a user-originated operation
+                        return true;
+                }
+
                 if (empty($token) || strpos($token, '|') === false) {
                     throw new Error('Invalid CSRF token provided');
                 }
--- a/src/Common/CsrfCounterMeasure.php
+++ b/src/Common/CsrfCounterMeasure.php
@@ -41,14 +41,35 @@ trait CsrfCounterMeasure
     /**
      * Create a form element to countermeasure CSRF attacks
      *
+     * If the {@see requestIsSafe()} check concludes the request is safe, returns a dummy element that accepts any
+     * value. If it concludes the request is unsafe, throws an {@see Error}. If the check is inconclusive (legacy
+     * browser), creates a token-based CSRF element validated against $uniqueId.
+     *
      * @param string $uniqueId A unique ID that persists through different requests
      *
      * @return FormElement
+     *
+     * @throws Error If {@see requestIsSafe()} returns false
      *
      * @deprecated Use {@see addCsrfCounterMeasure()} instead
      */
     protected function createCsrfCounterMeasure($uniqueId)
     {
+        $requestIsSafe = $this->requestIsSafe();
+
+        if ($requestIsSafe !== null) {
+            if (! $requestIsSafe) {
+                throw new Error('Rejecting cross-site request');
+            }
+
+            return new HiddenElement('CSRFToken', [
+                'ignore'     => true,
+                'validators' => ['Callback' => function () {
+                    return true;
+                }]
+            ]);
+        }
+
         $hashAlgo = in_array('sha3-256', hash_algos(), true) ? 'sha3-256' : 'sha256';
 
         $seed = random_bytes(16);
@@ -107,5 +128,24 @@ trait CsrfCounterMeasure
         }
 
         $this->addElement($this->createCsrfCounterMeasure($uniqueId ?? $this->csrfCounterMeasureId));
+    }
+
+    /**
+     * Get whether the request is safe from CSRF based on the Sec-Fetch-Site header
+     *
+     * Returns true if the header indicates a same-origin request or a user-originated operation (e.g. typing a URL),
+     * both of which cannot be forged by a cross-site attacker. Returns null if the header is absent,
+     * in which case a CSRF token must be validated. False indicates a cross-site request.
+     *
+     * @return ?bool
+     */
+    protected function requestIsSafe(): ?bool
+    {
+        return match ($_SERVER['HTTP_SEC_FETCH_SITE'] ?? null) {
+            'same-origin' => true, // same scheme, host and port
+            'none'        => true, // a user-originated operation
+            null          => null, // legacy browser without Sec-Fetch-Site support
+            default       => false,
+        };
     }
 }
--- a/tests/Common/CsrfCounterMeasureTest.php
+++ b/tests/Common/CsrfCounterMeasureTest.php
@@ -8,14 +8,28 @@ use ipl\Html\Form;
 use ipl\Html\FormElement\HiddenElement;
 use ipl\Tests\Web\TestCase;
 use ipl\Web\Common\CsrfCounterMeasure;
+use PHPUnit\Framework\Attributes\DataProvider;
 
 class CsrfCounterMeasureTest extends TestCase
 {
+    protected function setUp(): void
+    {
+        parent::setUp();
+        unset($_SERVER['HTTP_SEC_FETCH_SITE']);
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+        unset($_SERVER['HTTP_SEC_FETCH_SITE']);
+    }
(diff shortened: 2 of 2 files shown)


sukhwinder33445 (approved):

src/Common/CsrfCounterMeasure.php
```diff
@@ -108,4 +112,13 @@ protected function addCsrfCounterMeasure(?string $uniqueId = null): void
 
         $this->addElement($this->createCsrfCounterMeasure($uniqueId ?? $this->csrfCounterMeasureId));
     }
+
+    protected function requestIsSafe(): bool
```
TheSyscall: Missing docstring

src/Common/CsrfCounterMeasure.php
```diff
@@ -72,6 +72,10 @@ protected function createCsrfCounterMeasure($uniqueId)
             }]
         ];
 
+        if ($this->requestIsSafe()) {
+            unset($options['validators']);
+        }
+
```
TheSyscall: Why not include this check as part of the validator?
It should only run when the form is actually sent. `addCsrfCounterMeasure` and `createCsrfCounterMeasure` are called from within `assemble`, so with any request type.
Adding the full CSRF token with validator in every request unconditionally makes sense for consistency, even if we end up ignoring its value later.

src/Common/CsrfCounterMeasure.php
```diff
@@ -98,7 +102,7 @@ public function hasValue(): bool
      */
     protected function addCsrfCounterMeasure(?string $uniqueId = null): void
     {
-        if (! $this->csrfCounterMeasureEnabled) {
+        if (! $this->csrfCounterMeasureEnabled || $this->requestIsSafe()) {
```
TheSyscall: See above

TheSyscall (changes requested):

nilmerg (changes requested):
This misses an important point, to me. Either that or it's not documented/mentioned at all.

The token was always only a mitigation for CSRF, hence the name CSRFToken. It does **not** authenticate/authorize. So by supporting this request header, we fully trust the browser. But why is there still a token in case the header value is `same-site` or `cross-site`? Why doesn't it simply fail?

If the header is set the browser supports it and sets it reliably with every request. Or is this not guaranteed? With the header set, the initial response can decide whether a token is required or not and any subsequent request either includes the header again or a token.

There's just no reality I see that a form is being submitted *legitimately* if the header says it's not `same-origin`.

src/Common/CsrfCounterMeasure.php
Al2Klimov: That's on purpose! I want to make setting the token optional.

src/Common/CsrfCounterMeasure.php
```diff
@@ -69,8 +74,12 @@ protected function createCsrfCounterMeasure($uniqueId)
                 }
 
                 return true;
-            }]
-        ];
+            }];
+        } else if (! $requestIsSafe) {
+            $options['validators'] = ['Callback' => function () {
+                throw new Error('Rejecting cross-site request');
```
nilmerg: `createCsrfCounterMeasure` is deprecated. There's no reason to return an element in case the request isn't considered safe. Neither does `addCsrfCounterMeasure` and throws immediately.

So instead of refactoring the entire method, let it early return or throw as well.

nilmerg (changes requested):
* Please also adjust the accompanying test

Another important point to clarify:

The rationale for solely relying on this header instead of still requiring a token is that we only care about real browsers. Clients such as cURL will be able to spoof this header but in reality, they never needed to provide a token in the first place as if the route in question supports direct access like this, transmitting it as `Content-Type: application/json` does not require it anyway. Any route not supporting this still requires the user to authenticate and this then means the client needs to posess a session cookie and is able to also forge a CSRF token on its own.

src/Common/CsrfCounterMeasure.php
Al2Klimov: > instead of refactoring the entire method, let it early return

Would this be acceptable for a safe request?

```php
return new HiddenElement('CSRFToken', ['ignore' => true]);
```

nilmerg: No, by looking at users of the trait I noticed one that expects this to return an element with a validator set that verifies a manually set token. Quite hacky, yes, but nonetheless, the method `create…` must still return an element and that should have at least a validator that succeeds.

Al2Klimov: Better?

```php
return new HiddenElement('CSRFToken', [
    'ignore'     => true,
    'validators' => ['Callback' => function () {
        return true;
    }]
]);
```

nilmerg: Obviously? Now start committing please…

src/Common/CsrfCounterMeasure.php
```diff
@@ -102,10 +117,39 @@ protected function addCsrfCounterMeasure(?string $uniqueId = null): void
...
+        $requestIsSafe = $this->requestIsSafe();
+
+        if ($requestIsSafe !== null) {
+            if (! $requestIsSafe) {
+                throw new Error('Rejecting cross-site request');
+            }
+
+            return;
+        }
+
```
TheSyscall: Why is this checked twice? Once in here and then again in `createCsrfCounterMeasure`.

As far as I can see skips only the check for the `csrfCounterMeasureId`, which should already be present for the form to be assembled anyway.

src/Common/CsrfCounterMeasure.php
TheSyscall, line 113: The request is safe behavior should be documented.

TheSyscall, line 48: The request is safe behavior should be documented.

src/Common/CsrfCounterMeasure.php
```diff
@@ -102,10 +117,39 @@ protected function addCsrfCounterMeasure(?string $uniqueId = null): void
...
         }
 
         $this->addElement($this->createCsrfCounterMeasure($uniqueId ?? $this->csrfCounterMeasureId));
     }
+
+    /**
+     * Get whether the request is safe from CSRF based on the Sec-Fetch-Site header
+     *
+     * Returns true if the header indicates a same-origin request or a user-originated operation (e.g. typing a URL),
+     * both of which cannot be forged by a cross-site attacker. Returns null if the header is absent,
```
TheSyscall, line 137: Maybe add clarification that it can't be forged with JavaScript by a browser

TheSyscall (changes requested):

src/Common/CsrfCounterMeasure.php
Al2Klimov: That's on purpose to allow omitting setCsrfCounterMeasureId(). After all, we also omit token generation now for new browsers.

src/Common/CsrfCounterMeasure.php
Al2Klimov: Well... that's what I say here?

src/Common/CsrfCounterMeasure.php
TheSyscall: This would lead to users on older browsers causing Errors/LogicExceptions #386 on the server, because `setCsrfCounterMeasureId()` was never called in the expectation that every request to this form can only come from new browsers.

We essentially set up Csrf protection with the expectation to never generate a CSRF-Token.

Al2Klimov: I made the message more clear.

TheSyscall: I'm still not quite happy with this code, it hides a potential implementation error behind an expectation on the users' browser.

@nilmerg What do you say to this?

nilmerg: @TheSyscall is right, I say. The error below signals a programming error (and should have used LogicException tbh) and in case said programmer uses a modern browser, his/her/its user with a legacy browser runs into it.

Al2Klimov: Done.

TheSyscall (approved):
👍

nilmerg (approved):
I verified how a browser handles `Sec-Fetch-Site: none` in case of service workers, as I was curios whether it is okay that this change concludes that the request is safe in this case. For this, I let a service-worker control every network request and monitored which value is sent. The observation I made was that until the request is transmitted using the worker, no Sec-Fetch-Site header is part of the request. Only when the worker transmits it, the browser detects this as programmatic fetch, the value is set to `same-origin` or `cross-origin`.

So my paranoia-me is satisfied.

PR #306, opened 2025-08-06
Align title and caption text in minimal layout

jrauh01:
## Problem
In the minimal view mode in item lists, the check output is not properly aligned vertically with the remaining text, it's baseline is slightly too high. This is because the header is a **flexbox** with `align-items: flex-start`. Setting `align-items: baseline` did increase the total `height` of the header from `18px` to `19px`. This is not what we want and would affect the control layout in the detail area of an object.

## Solution
To solve this we have to set the header's height to fixed `1.5em`. The `align-items: baseline` now aligns the texts without increasing the `height`.

Fixes https://github.com/Icinga/icingadb-web/issues/785

### DIFF
--- a/asset/css/item-layout.less
+++ b/asset/css/item-layout.less
@@ -50,11 +50,12 @@
 
   header {
     display: flex;
-    align-items: flex-start;
+    align-items: baseline;
     justify-content: space-between;
   }
   &.minimal-item-layout > .main > header {
     max-width: 100%;
+    height: 1.5em;
   }
 
   .caption {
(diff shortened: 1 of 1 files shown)


asset/css/item-layout.less
```diff
@@ -50,8 +50,10 @@
 
   header {
     display: flex;
-    align-items: flex-start;
+    align-items: baseline;
     justify-content: space-between;
+
+    height: 1.5em;
```
nilmerg, line 56: This must not be applied to `detailed-item-layout`. The header is allowed to wrap in this case.

nilmerg (changes requested):

PR #311, opened 2025-08-22
CompatForm: Use new decorators

sukhwinder33445:


src/Compat/FormDecorator/CheckboxDecorator.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        $results
+            ->append($formElement)
+            ->skipDecorators('RenderElement')
+            ->append(new HtmlElement(
+                'label',
+                $attributes,
+                new HtmlElement('span', Attributes::create(['class' => 'toggle-slider']))
+            ));
```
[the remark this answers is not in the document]
nilmerg: It doesn't play nice with already existing forms otherwise.

src/Compat/CompatForm.php
```diff
@@ -3,21 +3,79 @@
...
+                'name' => 'HtmlTag',
+                'options' => [
+                    'tag' => 'div',
+                    'class' => 'control-label-group',
+                    'condition' => function (FormElement $element): bool {
+                        return ! $element instanceof FormSubmitElement
+                            && (! $element instanceof HtmlElementInterface || $element->getTag() !== 'fieldset');
+                    }
+                ]
+            ],
```
nilmerg, line 55: Do this with a custom label decorator instead. If there's no label, it must default to use `&nbsp;`

src/Compat/FormDecorator/LabelDecorator.php
```diff
@@ -0,0 +1,46 @@
...
+use ipl\Html\Attributes;
+use ipl\Html\Contract\FormElement;
+use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\Contract\HtmlElementInterface;
+use ipl\Html\FormDecorator\DecorationResults;
+use ipl\Html\HtmlElement;
+use ipl\Html\HtmlString;
+use ipl\Html\Text;
+use ipl\Html\FormDecorator\LabelDecorator as IplHtmlLabelDecorator;
+/**
```
nilmerg, line 14: missing empty lijne

src/Compat/FormDecorator/LabelDecorator.php
```diff
@@ -0,0 +1,46 @@
...
+ * Decorates the label of the form element
+ */
+class LabelDecorator extends IplHtmlLabelDecorator
```
nilmerg, line 17: At the moment, extending this class does not have any advantage. Please give it this one:

ipl\Html should provide a method to retrieve the form element label: `protected function getElementLabel(FormElement $formElement): ?ValidHtml`

Then this decorator can override it with:

```
        $label = parent::getElementLabel($formElement);
        if ($label === null) {
            return HtmlString::create('&nbsp;');
        }
```

src/Compat/FormDecorator/LabelDecorator.php
```diff
@@ -0,0 +1,46 @@
...
+            $labelAttr = null;
+            if ($isHtmlElement && $formElement->getAttributes()->has('id')) {
+                $labelAttr = new Attributes(['for' => $formElement->getAttributes()->get('id')->getValue()]);
+            }
+
+            $wrapper->addHtml(new HtmlElement('label', $labelAttr, new Text($formElement->getLabel())));
+        }
+
+        $results->prepend($wrapper);
+    }
```
nilmerg, line 45: drop this

nilmerg (changes requested):

PR #316, opened 2025-09-29
TermInput: Sanitize container IDs properly

### LINKED ISSUE #315: `TermInput` broken if it's name contains CSS selector operators
The `\ipl\Web\FormElement\TermInput` element uses its name to assign its containers an ID. With the name `my.name`, the browser evaluates this as `#my.name` requiring the ID `my` and a class `name`, instead of an ID `my.name`.

Additionally, if used as part of a fieldset, the name used is invalid or might be a duplicate as the fieldset's assigned name is ignored.

nilmerg:
fixes #315 

requires https://github.com/Icinga/ipl-html/pull/171

### DIFF
--- a/src/FormElement/TermInput.php
+++ b/src/FormElement/TermInput.php
@@ -2,6 +2,7 @@
 
 namespace ipl\Web\FormElement;
 
+use ipl\Html\Attribute;
 use ipl\Html\Attributes;
 use ipl\Html\Form;
 use ipl\Html\FormElement\FieldsetElement;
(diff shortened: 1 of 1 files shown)


src/FormElement/TermInput.php
```diff
@@ -268,19 +269,19 @@ public function prepareMultipartUpdate(ServerRequestInterface $request): array
         if ($this->valueHasBeenPasted()) {
             $updates[] = $this->termContainer();
             $updates[] = [
-                HtmlString::create(json_encode(['#' . $this->getName() . '-search-input', []])),
+                HtmlString::create(json_encode(['#' . Attribute::sanitizeId($this->getName()) . '-search-input', []])),
```
raviks789, line 272: This might still not address the duplicate ID issue, if the term inputs are added to different fieldsets with the same name in the same form. In this case both the term inputs receives the same ID. That was the reason I had previously used `getValueOfNameAttribute` instead.

raviks789 (approved):
LGTM!

PR #319, opened 2025-10-14
Let `LabelDecorator` mark required elements as such with an asterisk

### LINKED ISSUE #239: Mark required form elements with an asterisk
`Icinga\Web\Form` does this already:

image

> The [new decorator implementation](https://github.com/Icinga/ipl-html/blob/main/doc/45-Form-Decorators.md) allows to easily implement this. `\ipl\Web\Compat\FormDecorator\LabelDecorator` is a good basis and should be extended to render the asterisk as [@flourish86](https://github.com/flourish86) suggested above. In addition, it should implement `\ipl\Html\Contract\FormDecoration` to track if there is any required element in the form and render a permanently visible description at the bottom below the submit buttons. Take a look at some legacy forms in Icinga Web to get a feel for how this should look. (e.g., General configuration)

BastianLedererIcinga:
Add implementation of LabelDecorator that marks required fields with tooltops and creates an explanation below the form if necessary.

resolves #239

### DIFF
--- a/tests/Compat/CompatFormTest.php
+++ b/tests/Compat/CompatFormTest.php
@@ -157,20 +157,27 @@ HTML;
     public function testLabelDecoration(): void
     {
         $this->form->applyDefaultElementDecorators()
-            ->addElement('text', 'test_text_non_required', ['required' => false, 'label' => 'test_non_required'])
+            ->addElement(
+                'text',
+                'test_text_non_required',
+                ['required' => false, 'label' => 'test_non_required', 'id' => 'test-id-required']
+            )
             ->addElement('text', 'test_text_no_label')
-            ->addElement('text', 'test_text_required', ['required' => true, 'label' => 'test_required'])->render();
+            ->addElement(
+                'text',
+                'test_text_required',
+                ['required' => true, 'label' => 'test_required', 'id' => 'test-id-non-required']
+            );
 
-        $expected = sprintf(
-            <<<'HTML'
+        $expected = <<<'HTML'
     <form class="icinga-form icinga-controls" method="POST">
         <div class="control-group">
             <div class="control-label-group">
-                <label class="form-element-label" for="%s">
+                <label class="form-element-label" for="test-id-required">
                     test_non_required
                 </label>
             </div>
-            <input name="test_text_non_required" type="text" id="%s"/>
+            <input name="test_text_non_required" type="text" id="test-id-required"/>
         </div>
         <div class="control-group">
             <div class="control-label-group">
@@ -180,12 +187,12 @@ HTML;
         </div>
         <div class="control-group">
             <div class="control-label-group">
-                <label class="form-element-label" for="%s">
+                <label class="form-element-label" for="test-id-non-required">
                     test_required
                     <span class="required-hint" aria-hidden="true" title="Required"> *</span>
                 </label>
             </div>
-            <input required aria-required="true" name="test_text_required" type="text" id="%s"/>
+            <input required aria-required="true" name="test_text_required" type="text" id="test-id-non-required"/>
         </div>
         <ul class="form-info">
             <li>
@@ -193,12 +200,7 @@ HTML;
             </li>
         </ul>
     </form>
-HTML,
-            $this->form->getElement('test_text_non_required')->getAttribute('id')->getValue(),
-            $this->form->getElement('test_text_non_required')->getAttribute('id')->getValue(),
-            $this->form->getElement('test_text_required')->getAttribute('id')->getValue(),
-            $this->form->getElement('test_text_required')->getAttribute('id')->getValue(),
-        );
+HTML;
 
         $this->assertHtml($expected, $this->form);
     }
--- a/tests/Compat/FormDecorator/LabelDecoratorTest.php
+++ b/tests/Compat/FormDecorator/LabelDecoratorTest.php
@@ -32,19 +32,17 @@ class LabelDecoratorTest extends IplHtmlTestCase
     {
         $formElement = new TextElement('test', [
             'required' => true,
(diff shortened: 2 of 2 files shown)


asset/css/compat.less
```diff
@@ -25,6 +25,15 @@ form.icinga-form {
   }
 }
 
+.icinga-controls {
+  .required-hint {
+    font-size: 1.2em;
```
nilmerg: Please don't change the font size. This results in a higher than normal line height and offsets most of the form elements.

src/Compat/FormDecorator/LabelDecorator.php
```diff
@@ -6,75 +6,60 @@
...
-        if (! $elementAttrs->has('id')) {
-            $elementAttrs->set('id', uniqid('form-element-'));
-        }
-
-        $result = new HtmlElement(
-            'label',
-            null,
-            HtmlString::create($label)
-        );
+        $result = parent::getElementLabel($formElement) ?? HtmlString::create('&nbsp;');
```
nilmerg: Not sure if it's a good idea to render something like that: `&nbsp; <span…`

Please return early with the non-breaking space.

src/Compat/FormDecorator/LabelDecorator.php
```diff
@@ -2,15 +2,67 @@
...
+use ipl\Html\HtmlElement;
+use ipl\I18n\Translation;
+use ipl\html\Text;
 
-class LabelDecorator extends IplHtmlLabelDecorator
+class LabelDecorator extends IplHtmlLabelDecorator implements FormDecoration
 {
+    use Translation;
+
+    protected bool $requiredExplanationNeeded = false;
```
nilmerg: Add a short docstring please: `/** @var … */`

tests/Compat/FormDecorator/LabelDecoratorTest.php
```diff
@@ -23,4 +27,52 @@ public function testWithoutLabelAttribute(): void
...
+
+    public function testWithRequiredField(): void
+    {
+        $formElement = new TextElement('test', [
+            'required' => true,
+            'label' => 'test-label'
+        ]);
+
+        $results = new FormElementDecorationResult();
+        $form = (new CompatForm())->addElement($formElement);
```
nilmerg: Unittests should be lightweight and not depend on unrelated components or implementation details. You are instantiating a form only, because `decorateForm` requires one. Instead, please create and use a stub or mock.

tests/Compat/FormDecorator/LabelDecoratorTest.php
```diff
@@ -23,4 +27,52 @@ public function testWithoutLabelAttribute(): void
 
         $this->assertSame('&nbsp;', $results->assemble()->render());
     }
+
+    public function testWithRequiredField(): void
```
nilmerg: You should include an additional case:

- decorate a required elemement
- decorate a non-required element
- decorate the form
- assert that the form-info is rendered

We're dealing with internal state after all, so let's make sure it works as designed :)

tests/Compat/FormDecorator/LabelDecoratorTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace ipl\Tests\Web\Compat\FormDecorator;
 
+use ipl\Html\Form;
```
nilmerg: `ipl\Html\Contract\Form` please

tests/Compat/FormDecorator/LabelDecoratorTest.php
```diff
@@ -23,4 +28,74 @@ public function testWithoutLabelAttribute(): void
...
+            'label' => 'test-label'
+        ]);
+
+        $results = new FormElementDecorationResult();
+        $formStub = $this->createStub(Form::class);
+
+        $this->decorator->decorateFormElement($results, $formElement);
+        $this->decorator->decorateForm($results, $formStub);
+
+        $this->assertStringNotContainsString(
```
nilmerg: As this represents the counterpart of `testFormWithRequiredElement`, what about using `assertStringEndsNotWith`? :P

nilmerg (changes requested):

tests/Compat/FormDecorator/LabelDecoratorTest.php
```diff
@@ -23,4 +27,95 @@ public function testWithoutLabelAttribute(): void
...
+
+        $results = new FormElementDecorationResult();
+        $this->decorator->decorateFormElement($results, $formElement);
+
+        $this->assertStringNotContainsString(
+            '<span class="required-hint" aria-hidden="true" title="Required"> *</span>',
+            $results->assemble()->render()
+        );
+        $this->assertTrue(! $formElement->hasAttribute('aria-required')
+            || $formElement->getAttribute('aria-required')->getValue() !== "true");
```
nilmerg: Please simplify this :laughing: `assertFalse` and check only for the attribute's existence.

tests/Compat/CompatFormTest.php
```diff
@@ -150,4 +153,53 @@ public function testDuplicateSubmitButtonRespectsOriginalAttributes(): void
...
+            <li>
+                * Required field
+            </li>
+        </ul>
+    </form>
+HTML,
+            $this->form->getElement('test_text_non_required')->getAttribute('id')->getValue(),
+            $this->form->getElement('test_text_non_required')->getAttribute('id')->getValue(),
+            $this->form->getElement('test_text_required')->getAttribute('id')->getValue(),
+            $this->form->getElement('test_text_required')->getAttribute('id')->getValue(),
```
nilmerg: Same here, use explicit IDs please

Issue #331, opened 2025-11-05
FilterEditor: Cannot clear filter


sukhwinder33445:
### To reproduce:
1.	Create a filter via the search bar.
2.	Try to remove the filter via the `FilterEditor`.

### Observed behavior:
- The `FilterEditor` cannot be applied when a filter is removed (via the `delete icon`), because the input field is `required`.
- Clicking the `Clear Filter` button does not clear the filter properly, `?=` remains in the URL and an error is thrown.

nilmerg:
> The FilterEditor cannot be applied when a filter is removed (via the delete icon), because the input field is required.

If you click on the delete icon for the accompanying chain the clear button appears and we're back at the second observation 🤣 

--

Well, the clear button should probably be always visible. And always work fine.

PR #332, opened 2025-11-10
Fix/clear filter

BastianLedererIcinga:
fix: #331 
The `Clear Filter` button is now always available, and after using it no `?=` will be in the url.


sukhwinder33445 (commented):
LGTM
Please squash the commits.

nilmerg (changes requested):

nilmerg (changes requested):

nilmerg (approved):

PR #336, opened 2025-11-12
Introduce `PrimaryButtonDecorator`

sukhwinder33445:


src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -0,0 +1,33 @@
...
+
+            $classes = implode(' ', (array) $element->getAttributes()->get('class')->getValue());
+
+            if (preg_match('/(?<=^|\s)btn-primary(?=$|\s)/', $classes)) {
+                return;
+            }
+
+            if (! $primaryButtonCandidate && ! preg_match('/\bbtn-[\w-]+/', $classes)) {
+                $primaryButtonCandidate = $element;
+            }
```
nilmerg: Can't you just cast the value to an array and handle it as such? I think it's bad practice to define multiple classes as one single string anyway, exactly due to this reason. It would make searching for other `btn-*` classes also easier.

Have you found cases where button classes are defined in a string with other classes or are you only anticipating such?

src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -0,0 +1,42 @@
...
+
+            /** @var string[] $classes */
+            $classes = (array) $element->getAttributes()->get('class')->getValue();
+
+            // If a class attribute's value is defined as a string such as 'a b c' or as an array like ['a b c'],
+            // it is not checked for existing 'btn-*' or 'btn-primary' classes.
+            if (count($classes) === 1 && str_contains($classes[0], ' ')) {
+                $primaryButtonCandidate = $element;
+                break;
+            }
```
nilmerg: Ehm, not sure this "optimization" is necessary.

src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -0,0 +1,42 @@
...
+
+            if (in_array('btn-primary', $classes)) {
```
nilmerg: strict: true

src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -0,0 +1,42 @@
...
+                return;
+            }
+
+            if (! $primaryButtonCandidate && ! array_find($classes, fn($v) => str_starts_with($v, 'btn-'))) {
+                $primaryButtonCandidate = $element;
```
nilmerg: break?

src/Compat/FormDecorator/PrimaryButtonDecorator.php
sukhwinder33445: Without this condition, a value starting with `btn-a test` will skip the final if check because of the expression `! array_find($classes, fn($v) => str_starts_with($v, 'btn-'))`, and the class is not added.

src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -0,0 +1,42 @@
...
+    public function decorateForm(DecorationResult $result, Form $form): void
+    {
+        $primaryButtonCandidate = null;
+        foreach ($form->getElements() as $element) {
+            if (! $element instanceof FormSubmitElement) {
+                continue;
+            }
+
+            /** @var string[] $classes */
+            $classes = (array) $element->getAttributes()->get('class')->getValue();
```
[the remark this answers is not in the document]
sukhwinder33445: Get creates one with `null` as value.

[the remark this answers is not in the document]
nilmerg: Exactly. The array contains null then, not just strings, so your type hint is incorrect and the remaining routine's behavior is undefined.

PR #350, opened 2026-01-08
Decouple time widgets

### LINKED ISSUE #341: Decouple `TimeAgo`, `TimeUntil` and `TimeSince` from Icinga Web
All three widgets rely at the moment on Icinga Web:

* `\Icinga\Date\DateFormatter` to format datetime values
* https://github.com/Icinga/icingaweb2/blob/70f29827f921f24b2afc0556c4800dfa55a2d0c1/public/js/icinga/ui.js#L484 to animate them in the UI

To get rid of `\Icinga\Date\DateFormatter::formatDateTime()` is relatively straightforward since [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time#valid_datetime_values) elements mandate a specific format that can be hard-coded.

`\Icinga\Date\DateFormatter::timeAgo()` (and all its counterparts) on the other hand may be simply copied over to the respective widget.

Though, all three of them are based on `\Icinga\Date\DateFormatter::diff()` so why not introducing a new widget: `ipl\Web\Widget\Time`. The other three extend it and can access its protected method `ipl\Web\Widget\Time::diff()`. (Which is a modernized copy of `\Icinga\Date\DateFormatter::diff()`, also consider using `\DateTime::diff()`)

Note that I don't think `ipl\Web\Widget\Time` should be abstract, it should rather provide a reasonable default content. (e.g. not declare the `datetime` attribute and instead rendering the datetime in the content)
One might also think about `ipl\Web\Widget\Time::relative(\DateTime): static` that automatically chooses `TimeUntil` or `TimeAgo` depending on the input.

For the client side, we should provide a default Icinga Web behavior called `RelativeTime` that provides an implementation very similar to https://github.com/Icinga/icingaweb2/blob/70f29827f921f24b2afc0556c4800dfa55a2d0c1/public/js/icinga/ui.js#L484 but using ECMAScript 2015 features only. (No jQuery) Instead of reacting on CSS classes, it should also react on data attributes.

Talking of modernized JS, don't forget that there's [`Intl.RelativeTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) which might serve as alternative to the server side localization.

Jan-Schuppik:
This PR removes the direct dependency of the time widgets on Icinga Web (\Icinga\Date\DateFormatter::formatDateTime()) and introduces a native relative-time formatting behavior.

**Key differences to the old Icinga Web implementation**

Previously the following markup was animated:

```html
<span class="relative-time time-ago" title="2026-04-21 14:44:34">3m4s ago</span>
```

The new implementation still supports this, but favors the following new markup:

```html
<time class="time-ago" data-relative-time="ago" datetime="2026-04-21T14:44:34" title="2026-04-21 14:44:34">3m4s ago</time>
```

resolves #341


src/Widget/Time.php
```diff
@@ -0,0 +1,221 @@
...
+     * @var int
+     */
+    public const DATE = 2;
+
+    /**
+     * Format date and time
+     *
+     * @var int
+     */
+    public const DATETIME = 4;
```
sukhwinder33445: Please define oneline PHPDoc, as it looks clearer.

src/Widget/Time.php
```diff
@@ -0,0 +1,221 @@
...
+
+    /** @var DateTimeImmutable */
+    protected DateTimeImmutable $time;
+
+    /** @var string */
+    protected string $dateTime;
+
+    protected $tag = 'time';
+
+    public function __construct(int|float|DateTimeInterface $time)
```
sukhwinder33445: Please add PHPDoc.

src/Widget/Time.php
```diff
@@ -0,0 +1,206 @@
...
+     * Assemble specific logic of subclasses
+     * Override this method to customize the output
+     *
+     * @return void
+     */
+    protected function assembleSpecific(): void
+    {
+        $this->addAttributes(Attributes::create(['data-relative-time' => null]));
+        $this->add($this->dateTime);
+    }
```
nilmerg: Drop `assembleSpecific`. Don't set nor reset `data-relative-time`. Set the `title` already in the constructor. The default implementation of `assemble` then just adds the datetime as content, but by using `addHtml(Text::create`.

src/Widget/Time.php
```diff
@@ -0,0 +1,206 @@
...
+        if ($value instanceof DateTimeImmutable) {
+            return $value;
+        }
+
+        if ($value instanceof DateTimeInterface) {
+            return DateTimeImmutable::createFromInterface($value);
+        }
+
+        return new DateTimeImmutable('@' . (int) $value);
+    }
```
nilmerg: I see why this method exists, though why is it public? It's only used internally and that's because of one sole reason:

The class accepts a multitude of types as datetime. If that wouldn't be the case and it only accepts `\DateTime`, this method is obsolete. So my actual question is, why does this class accept so many types? Do we need this somewhere?

If not, drop it and require `\DateTime` everywhere and leave it untouched. (i.e. Don't change its timezone)

src/Widget/Time.php
nilmerg: Note that this comment is for `assemble` and `assembleSpecific`.

src/Widget/TimeAgo.php
```diff
@@ -2,32 +2,42 @@
...
+                $values = ['on %s', 'An event happened on the given date or date and time'];
+                break;
+            case static::RELATIVE:
+                $values = ['%s ago', 'An event that happened the given time interval ago'];
+                break;
+            case static::TIME:
+                $values = ['at %s', 'An event happened at the given time'];
+        }
+
+        return sprintf(t(...$values), $time);
```
nilmerg: https://github.com/Icinga/developer-guidelines-drafts/issues/10#issuecomment-3895818014

src/Widget/Time.php
```diff
@@ -0,0 +1,206 @@
...
+    ): array {
+        $time = static::toImmutable($time);
+        $compareTime = static::toImmutable($compareTime);
+
+        $interval = $compareTime->diff($time);
+
+        if ($interval->d > 2 || $interval->m > 0 || $interval->y > 0) {
+            $type = static::DATE;
+            $formatted = $time->format(date('Y') === date('Y', $time->getTimestamp()) ? 'M j' : 'Y-m');
+        } elseif ($interval->d > 0) {
```
nilmerg: Instead of `->d` please use `->days`, as I compared this with the previous implementation and didn't found a counterpart for `->m > 0` or `->y > 0` so I think you've added them because `->d` is not a total, whereas `->days` is.

nilmerg (changes requested):
Didn't look at JS yet.

Also, add tests for `Time` and its child classes. Test formatting and rendering.

nilmerg (changes requested):
There are still no tests!

nilmerg (changes requested):
I stopped reading the tests once I saw more than one skip due to when the test is performed. Sorry, but this just non-sense. Tests must **always** succeed.

In addition, they make too heavy usage of *contains* and *regex* assertions. They test a HTML widget, so should assert the HTML structure: `$obj` actual, `$html` expected. Other tests exactly work like this and there is even a predefined assertion as part of `ipl\Tests\Html\TestCase`.

So please refactor your tests accordingly.

asset/js/behavior/RelativeTimeBehavior.js
```diff
@@ -0,0 +1,34 @@
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | GPLv2+ */
```
BastianLedererIcinga: This is no longer part of Icinga Web 2.

asset/js/widget/RelativeTime.js
```diff
@@ -0,0 +1,149 @@
...
+
+            if (this._templateCache.has(type)) {
+                return this._templateCache.get(type);
+            }
+
+            const content = element.textContent || '';
+            let match = TIME_REGEX_FULL.exec(content) || TIME_REGEX_MIN_ONLY.exec(content);
+
+            const template = {
+                prefix: (match?.[1]).replace(/-+$/g, '') ?? '',
```
BastianLedererIcinga: If `match` is `null` `replace()` is called on `undefined`, so you need a second check if you want to be safe.

asset/js/widget/RelativeTime.js
```diff
@@ -0,0 +1,149 @@
...
+                const minutes = parseInt(partialTime[2], 10);
+                const seconds = parseInt(partialTime[3], 10);
+
+                let secondsDiff = (isNegative ? -1 : 1) * (minutes * 60 + seconds);
+
+                return future ? --secondsDiff : ++secondsDiff;
+            };
+
+            return fromTextContent(element, future);
+            // return fromDateTimeWithTimezone(element, future);
```
BastianLedererIcinga: If there are no plans to use this please remove it and the dead code referenced by it.

tests/Widget/TimeAgoTest.php
```diff
@@ -0,0 +1,186 @@
...
+        $this->assertSame($html, $rendered);
+
+        $timestampEventFloat = (float) $timestampEvent;
+        $timestampNowFloat = (float) $timestampNow;
+        $rendered2 = (new TimeAgo($timestampEventFloat, $timestampNowFloat))->render();
+
+        $this->assertSame($html, $rendered2);
+    }
+
+    public function testRenderHasAttributes(): void
```
BastianLedererIcinga: Please refactor the tests as @nilmerg suggested. This entire test, `testRenderHasAgoLabel` and `testRenderUsesTimeHtmlTag` become superfluous once `assertHtml` is used.

BastianLedererIcinga (commented):
I see my [icingadb-web PR](https://github.com/Icinga/icingadb-web/pull/1322) is still not entirely compatible with your changes, since I see no proper way to include a `data-ago-label` in `Time` widgets, which is required for the switch from absolute to relative. Let's discuss a solution and this time document the agreed approach in this PR.

src/Widget/TimeAgo.php
```diff
@@ -2,32 +2,60 @@
...
-    protected $defaultAttributes = ['class' => 'time-ago'];
-
-    public function __construct($ago)
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param int|float|DateTime|null $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, int|float|DateTime|null $compareTime = null)
```
nilmerg: `$compareTime` didn't exist beforehand, please only accept `?DateTime`

src/Widget/TimeSince.php
```diff
@@ -2,32 +2,51 @@
...
-    protected $defaultAttributes = ['class' => 'time-since'];
-
-    public function __construct($since)
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param int|float|DateTime|null $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, int|float|DateTime|null $compareTime = null)
```
nilmerg: Same here, constrain `$compareTime` to `?DateTime`

src/Widget/TimeUntil.php
```diff
@@ -2,33 +2,70 @@
...
-    protected $defaultAttributes = ['class' => 'time-until'];
-
-    public function __construct($until)
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param int|float|DateTime|null $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, int|float|DateTime|null $compareTime = null)
```
nilmerg: And here, `?DateTime $compareTime`

composer.json
```diff
@@ -26,7 +26,8 @@
     "ipl/scheduler": ">=0.1.0",
     "ipl/stdlib": ">=0.13.0",
     "fortawesome/font-awesome": "^6",
-    "wikimedia/less.php": "^3.2.1"
+    "wikimedia/less.php": "^3.2.1",
+    "ext-intl": "*"
```
nilmerg: Please move this up to the top where `ext-json` is, to also solve the conflict with `main`

asset/js/widget/RelativeTime.js
nilmerg, line 1: This isn't bound to a single element and as such doesn't fit into what a *widget* is in the sense how existing widgets work. Please move this one level up into `asset/js/`.

asset/js/widget/RelativeTime.js
```diff
@@ -0,0 +1,149 @@
+define([], function () {
```
nilmerg: Do not pass a requirements list if there are none.

asset/js/behavior/RelativeTimeBehavior.js
```diff
@@ -0,0 +1,34 @@
...
+
+            this._relativeTime.update(document);
+
+            if (this._timerHandle == null) {
+                this._timerHandle = icinga.timer.register(
+                    () => {this._relativeTime.update(document); },
+                    this,
+                    1000
+                );
+            }
```
nilmerg: I'd like to propose a different approach here, as this should only serve a single purpose: Bridging the gap between Icinga Web and this library.

This means that relying on Icinga Web's timer implementation for it work is an unnecessary dependency. What this behavior should be responsible for, is to abstract the knowledge when page content is available or updated. Updating the time elements regularly is a task for the library alone.

So my proposal is this:

The library JS should keep a registry of *known* elements. These are updated regularly by its own timer using `setInterval`. To know what elements still exist, it uses a `WeakMap` to keep a reference to them which disappear in case the elements disappear. To initially know about elements, it is told to scan upon specific events, which only this behavior can subscribe to in case of *Icinga Web*. But this isn't done using polling of course.

So once any elements are found the first time, the library starts the timer. Note that `WeakMap` isn't enumerable and so can only be used to detect whether a reference is still intact. A separate registry of known keys must therefore be used.

The behavior listens to `rendered` events and instructs the library to scan the newly rendered content only to update its set of elements.

It also overrides `Icinga.EventListener.unbind` to tell the library to stop the interval.

asset/js/widget/RelativeTime.js
nilmerg: Isn't the question why this implementation even exists to begin with? To me it makes the most sense after all, to not parse natural language but a machine-readable format. It's just rather useless right now considering what `_getTemplate` is used for: To re-construct the entire content from scratch. This easily breaks in case the content is translated to a language that doesn't fit the original structure. The only safe part to parse and re-construct is the time difference which always follows the format `\d+m \d+s`. So I'd favor parsing a machine-readable format and only animate the time difference and leave everything else alone.

nilmerg (changes requested):
@BastianLedererIcinga Please take over now, as @Jan-Schuppik is already occupied for the coming weeks and I want to finish this soon.

Don't hesitate to let Claude refactor the tests 😉

src/Widget/TimeAgo.php
```diff
@@ -2,32 +2,45 @@
...
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param ?DateTime $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, ?DateTime $compareTime = null)
+    {
+        if ($compareTime !== null) {
+            $this->compareTime = $this->castToDateTime($compareTime);
```
nilmerg: It doesn't need to be cast anymore ;)

src/Widget/TimeSince.php
```diff
@@ -2,32 +2,49 @@
...
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param ?DateTime $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, ?DateTime $compareTime = null)
+    {
+        if ($compareTime !== null) {
+            $this->compareTime = $this->castToDateTime($compareTime);
```
nilmerg: same here, no cast necessary

src/Widget/TimeUntil.php
```diff
@@ -2,33 +2,55 @@
...
+    /**
+     * @param int|float|DateTime|null $time Time as timestamp, DateTime object, or null for current time
+     * @param ?DateTime $compareTime Time to compare with, null for current time
+     *
+     * @throws Exception
+     */
+    public function __construct(int|float|DateTime|null $time = null, ?DateTime $compareTime = null)
+    {
+        if ($compareTime !== null) {
+            $this->compareTime = $this->castToDateTime($compareTime);
```
nilmerg: and here, you know the deal

asset/js/behavior/RelativeTimeBehavior.js
```diff
@@ -0,0 +1,34 @@
...
+            this.on('rendered', this.onTimeRendered, this);
+
+            /**
+             * RelativeTime instance
+             *
+             * @type {RelativeTime}
+             * @private
+             */
+            this._relativeTime = new RelativeTime(icinga.config.timezone);
+            this._relativeTime.scan(document);
```
nilmerg: I believe it shouldn't be necessary to scan upon construction. The `render` event is triggered upon page load by Icinga Web.

asset/js/behavior/RelativeTimeBehavior.js
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        onTimeRendered(event) {
+            event.data.self._relativeTime.scan(event.target);
```
nilmerg: Please only scan in case `event.currentTarget === event.target`. Otherwise this will scan containers multiple times while the event bubbles up to the ancestors.

For this to reliably work the listener registration in the constructor needs to use `.container` as condition.

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,115 @@
...
+
+            if (! this._timer) {
+                this._timer = setInterval(() => this.tick(), 1000);
+            }
+        }
+
+        tick() {
+            this._trackedElements.forEach((ref) => {
+                const el = ref.deref();
+                if (el) {
```
nilmerg, line 33: you only add to `_trackedElements`, drop dead refs from the set to avoid leaking memory.

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,115 @@
...
+
+        stop() {
+            if (this._timer !== null) {
+                clearInterval(this._timer);
+                this._timer = null;
+            }
+        }
+
+        updateElement(element) {
+            const DYNAMIC_RELATIVE_TIME_THRESHOLD = 60 * 60;
```
nilmerg: please make this a class constant/property

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,115 @@
...
+                element.textContent = element.textContent.replace(
+                    /-?\d+m \d+s/,
+                    this.render(remainingSeconds)
+                );
+            }
+        }
+
+        _getTimeDifferenceInSeconds(element, future = false) {
+
+            const fromDateTimeWithTimezone = (element, future = false) => {
```
nilmerg: Why is this a separate function?

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,115 @@
...
+                }
+
+                element.textContent = element.textContent.replace(
+                    /-?\d+m \d+s/,
+                    this.render(remainingSeconds)
+                );
+            }
+        }
+
+        _getTimeDifferenceInSeconds(element, future = false) {
```
nilmerg: please remove underscores from all methods

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,115 @@
...
+                const now = Date.now();
+
+                return Math.floor((future ? targetTimeUTC - now : now - targetTimeUTC) / 1000);
+            };
+
+            return fromDateTimeWithTimezone(element, future);
+        }
+
+        _getOffset() {
+            if (!this._offsetCache) {
```
nilmerg: a space after the negation operator please

nilmerg (changes requested):

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,116 @@
...
+            const offset = this.getOffset();
+
+            const targetTimeUTC = Date.parse(`${isoString}${offset}`);
+            const now = Date.now();
+
+            return Math.floor((future ? targetTimeUTC - now : now - targetTimeUTC) / 1000);
+        }
+
+        getOffset() {
+            if (! this._offsetCache) {
```
nilmerg: I discussed this with @Jan-Schuppik already. At this time I'm not sure a cached offset is actually of advantage as right now it has a severe limitation: Offset changes due to daylight saving reasons happen automatically and a browser tab that stays open during that time calculates time differences wrong after that. We may do X or Y, but honestly, let's just get rid of that cache.

asset/js/behavior/RelativeTimeBehavior.js
```diff
@@ -0,0 +1,35 @@
+define(["../RelativeTime", "icinga/legacy-app/Icinga"], function (RelativeTime, Icinga) {
```
nilmerg: I thought about backwards compatibility with Icinga Web < 2.14 recently. In this case, there's no `icinga/legacy-app/Icinga` and if loaded it collides with the old events.js implementation.

So I came up with an alternative loading mechanism:

```js
(function() {
    document.addEventListener('DOMContentLoaded', function() {
        …
    });
})();
```

In the constructor:

```js
                const RelativeTime = require('icinga/icinga-php-library/RelativeTime');

                // Icinga Web <= 2.13 support
                if (typeof icinga.ui.disableTimeCounters === 'function') {
                    this.on('icinga-init', null, () => icinga.ui.disableTimeCounters());
                }
```

nilmerg (changes requested):

asset/js/behavior/RelativeTimeBehavior.js
nilmerg: There's more to this. The new implementation requires `data-relative-time` which older versions of Icinga Web won't render. If we disable the old implementation as suggested, they won't be animated at all anymore. Instead, we have to still support e.g. `span.time-ago` in this case. So please let `RelativeTime` handle the old markup as well if we detect Icinga Web <= 2.13 here.

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,116 @@
...
+
+        stop() {
+            if (this._timer !== null) {
+                clearInterval(this._timer);
+                this._timer = null;
+            }
+        }
+
+        updateElement(element) {
+            const relativeTimeAgo = element.getAttribute('data-relative-time');
```
nilmerg: Just noticed this now: Please access data attributes using `.dataset` only.

asset/js/behavior/RelativeTimeBehavior.js
nilmerg: And even more: It's not only related to Icinga Web <= 2.13. This only relates to the animation implemenation, not the general usage. We have to support the old markup at all times.

* Expect `span` instead of `time` tags
* Infer the type of time from the class (e.g. `time-ago`)
* Infer the ISO datetime string from the `title` attribute

Please implement this in a way we can easily remove it sometime.

asset/js/RelativeTime.js
```diff
@@ -0,0 +1,130 @@
...
+        static DYNAMIC_RELATIVE_TIME_THRESHOLD = 60 * 60;
+
+        constructor(timezone) {
+            this.timezone = timezone;
+            this._trackedElements = new Set();
+            this._timer = null;
+        }
+
+        scan(root) {
+            const elements = root.querySelectorAll('time[data-relative-time], span.relative-time');
```
nilmerg: Sadly, `span.relative-time` does not suffice. There are custom implementations that only set `time-*` classes and the old implementation in ui.js didn't rely on `.relative-time` either.

nilmerg (changes requested):

nilmerg (approved):
Thank you both!

PR #355, opened 2026-03-04
Completer.js: Ensure autosubmit is triggered if no suggestion is selected

raviks789:
## Summary
If Completer has a `auto-submit` attribute in its dataset, Completer does not trigger autosubmit for manual inputs and the submit event is only triggered when a suggestion is clicked and if a classic `'class' => 'autosubmit'` attribute is used, the submit event will be triggered before the suggestion click and this in turn never fills the completer input with the suggestion data.

## Fix
Ensures the form submits automatically when:
- The completer is not part of term input
- The completer has the `auto-submit` dataset attribute
- No suggestion is explicitly selected on manual input

## Result
The solution triggers makes the completer to trigger autosubmit on manual input as expected when the focus goes away from the completer, even though no suggestion has been clicked. The classic autosubmit `'class' => 'autosubmit'` would still not work with the completer.

### DIFF
--- a/asset/js/widget/Completer.js
+++ b/asset/js/widget/Completer.js
@@ -478,7 +478,7 @@ define(["../notjQuery"], function ($) {
         }
 
         onFocusOut(event) {
-            // Autosubmit when the user leaves the input without selecting a suggestion on manually changing the value.
+            // Autosubmit if the user leaves the input and the input has been manually changed.
             // Only for non-instrumented mode — instrumented inputs (e.g. TermInput) handle
             // autosubmit themselves via BaseInput.autoSubmit() with proper term data.
             if (! this.instrumented && this.hasBeenManuallyChanged && this.shouldAutoSubmit()) {
@@ -486,7 +486,7 @@ define(["../notjQuery"], function ($) {
                 let input = event.target;
                 setTimeout(() => {
                     $(input.form).trigger('submit', { submittedBy: input });
-                }, 250);
+                }, 300);
             }
 
             if (this.completedInput === null) {
@@ -511,6 +511,7 @@ define(["../notjQuery"], function ($) {
                     }
 
                     this.hideSuggestions();
+                    this.hasBeenManuallyChanged = true;
                 }
             }, 250);
         }
@@ -668,6 +669,7 @@ define(["../notjQuery"], function ($) {
                     break;
                 case 'Tab':
                     suggestions = this.termSuggestions.querySelectorAll('[type="button"]');
+                    this.hasBeenManuallyChanged = false;
                     if (suggestions.length === 1) {
                         event.preventDefault();
                         let input = event.target;
@@ -689,6 +691,7 @@ define(["../notjQuery"], function ($) {
                     break;
                 case 'Escape':
                     if (this.hasSuggestions()) {
+                        this.hasBeenManuallyChanged = true;
                         this.hideSuggestions()
                         event.preventDefault();
                     }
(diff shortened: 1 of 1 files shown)


asset/js/widget/Completer.js
```diff
@@ -492,12 +494,20 @@ define(["../notjQuery"], function ($) {
                     && ! this.termSuggestions.contains(document.activeElement)
                 ) {
                     // Hide the suggestions if the user doesn't navigate them
-                    if (input !== completedInput) {
+                if (completedInput !== null && input !== completedInput) {
                         // Restore input if a suggestion lost focus
```
nilmerg: Leave this code path alone please. For one, the comment is not applicable anymore as it doesn't mention your change. Else, the remaining code wasn't intended to handle `completedInput = null` and you only adjusted one of three cases. Don't try to generalize code like this.

At the very start of the method:
- Check if `! this.instrumented && this.hasBeenManuallyChanged && this.shouldAutoSubmit()`
- Then immediately reset `this.hasBeenManuallyChanged`
- Now trigger the submit, either immediately or with its own timout

nilmerg (changes requested):
Is this still AI-generated? If not, remove the disclosure.

nilmerg (changes requested):
There's still an issue with duplicated auto submissions:
- Type in the input and wait for suggestions
- Choose a suggestion with the mouse
- Notice an aborted request in the network log

The request that is aborted should have never occurred because of the interaction with the suggestions.

This gets worse if you try to navigate the suggestion list with the keyboard.

> No, after these changes, it is not AI generated anymore. But don't we have to mention that it was initially AI generated and then adjusted (made necessary changes ) as I have added in the disclosure?

Once any trace of AI generated code is gone, there's no reason for the disclosure anymore. This is about for what AI was being used, nothing else. I'll write up my reasoning some time soon probably, so it's clear to everyone why.

nilmerg (changes requested):

nilmerg (changes requested):
So, I think it works better now than ever before. But I have a major complaint left, but I'm not sure if this can be mitigated at all as it may mean questioning the entire proposed change:

I can't stand it, in case JS is interfering with how I navigate through a UI. One key aspect to this is, that focus stays where I put it. Take the `SearchBar` or the `TermInput` (i.e. those who _instrument_ the completer). They both handle auto-submission in a way that doesn't affect the currently focused element too much, if at all. However, the completer will now while _non-instrumented_ auto-submit on focus-out, after 300ms. By this time the focus is long somewhere else, visibly, but then reverted back. Imagine a user typing already in some other field, or maybe an AI-Agent if you can't imagine someone (me 😅 ) typing so fast. (Though, no idea if an AI is really interacting this way with Web UIs…)

Anyway, this is only going to be used in Director right now, which already has a habit of being disgustingly bad in terms of focus handling. So users of Director won't notice this that much. 🙄

asset/js/widget/Completer.js
```diff
@@ -476,6 +483,23 @@ define(["../notjQuery"], function ($) {
...
         onFocusOut(event) {
+            setTimeout(() => {
+                // Autosubmit if the user leaves the input and the input has been manually changed.
+                // Only for non-instrumented mode — instrumented inputs (e.g. TermInput) handle
+                // autosubmit themselves via BaseInput.autoSubmit() with proper term data.
+                if (
+                    ! this.instrumented
+                    && this.hasBeenManuallyChanged
+                    && ! this.hasSuggestions()
+                    && this.shouldAutoSubmit()
```
nilmerg: `this.instrumented` and `this.shouldAutoSubmit()` are not influenced during runtime, so please do not push the callback onto the event stack for no reason.

nilmerg (changes requested):

PR #358, opened 2026-03-13
Introduce Callout

### LINKED ISSUE #349: Better form hints
Since this functionality has already been needed in several PRs, it would be great to introduce a **unified way to add well-designed hints to forms**. Centralizing the design would ensure **consistent appearance** across the application. Future changes would also apply to all usages.

Hints should be attachable either to the **entire form** or to **specific form elements**. We could also support different levels—such as `info`, `hint`, `warning`, or `error` - each with its own icon and color scheme.

The first example is from Icinga Notifications Web and could serve as a starting point. The design is clean and integrates well with the Icinga Web UI. In this case the hint is outside the form, but could be part of it.

Image

The following example is from the password policy PR from Icinga Web.

Image

ref https://github.com/Icinga/icingaweb2/pull/5430
ref https://github.com/Icinga/icingaweb2/pull/5419
ref https://github.com/Icinga/icinga-notifications-web/pull/369

TheSyscall:
Implement a callout box element that can be used to convey important information to the user.
This element is designed to be used above certain form elements, over the whole form or page.

It seems like the use of callout could conflict with the existing `FormDescription` and `FormNotification` decorators, at least within IW2 forms.
Callouts are not meant as a direct replacement for these decorators but as a more universal ipl widget that can be used even outside of forms to convey essential information to the user.

Merging these two concepts by extending the form-notification would be a viable alternative to this PR.

Callout types:
image

Callout full width:
image

Callout with optional title:
image

Callout above a form element:
image

closes #349

~requires Icinga/ipl-stdlib#77~

### DIFF
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -5,7 +5,7 @@
   column-gap: 1em;
 
   width: fit-content;
-  margin: 0 auto 1em auto;
+  margin: 0 auto 1em;
 
   &.fullwidth {
     width: 100%;
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -7,7 +7,7 @@
   width: fit-content;
   margin: 0 auto 1em;
 
-  &.fullwidth {
+  &.callout-fullwidth {
     width: 100%;
     justify-content: start;
   }
@@ -48,19 +48,19 @@
     font-size: 1.5em;
   }
 
-  &.info {
+  &.callout-type-info {
     --callout-color: @color-pending;
   }
 
-  &.success {
+  &.callout-type-success {
     --callout-color: @color-ok;
   }
 
-  &.warning {
+  &.callout-type-warning {
     --callout-color: @color-warning;
   }
 
-  &.error {
+  &.callout-type-error {
     --callout-color: @color-critical;
   }
 }
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -26,6 +26,10 @@
     margin: 0;
   }
 
+  .callout-title {
+    margin-bottom: .5em;
+  }
+
   .callout-text {
     display: flex;
     flex-direction: column;
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -9,6 +9,13 @@
 
   &.fullwidth {
     width: 100%;
+    justify-content: start;
+  }
+
+  &.form-callout {
+    margin-left: 14em;
+    width: auto;
+    justify-content: start;
   }
 
   i.icon::before {
@@ -22,11 +29,6 @@
   .callout-text {
     display: flex;
     flex-direction: column;
-  }
-
-  &.form-callout {
-    margin-left: 14em;
-    width: auto;
   }
 }
 
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -1,0 +1,60 @@
+// Layout
+.callout {
+  display: flex;
+  justify-content: center;
+  column-gap: 1em;
+
+  width: fit-content;
+  margin: 0 auto 1em auto;
+
+  &.fullwidth {
+    width: 100%;
+  }
+
+  i.icon::before {
+    margin-right: 0;
+  }
+
+  p {
+    margin: 0;
+  }
+
+  .callout-text {
+    display: flex;
+    flex-direction: column;
+  }
+
+  &.form-callout {
+    margin-left: 14em;
+    width: auto;
+  }
+}
+
+// Style
+.callout {
+  padding: .5em 1em;
+  border: 1px solid var(--callout-color);
+  background-color: color-mix(in srgb, var(--callout-color) 10%, transparent);
+  border-radius: .25em;
+
+  i.icon {
+    color: var(--callout-color);
+    font-size: 1.5em;
+  }
+
+  &.info {
+    --callout-color: @color-pending;
+  }
+
+  &.success {
+    --callout-color: @color-ok;
+  }
+
+  &.warning {
+    --callout-color: @color-warning;
+  }
+
+  &.error {
+    --callout-color: @color-critical;
+  }
+}
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -7,12 +7,12 @@
   width: fit-content;
   margin: 0 auto 1em;
 
-  &.callout-fullwidth {
+  &.callout-full-width {
     width: 100%;
     justify-content: start;
   }
 
-  &.form-callout {
+  &.callout-form-element {
     margin-left: 14em;
     width: auto;
     justify-content: start;
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -13,6 +13,11 @@
   }
 
   &.callout-form-element {
+    /*
+     * 14em is the width of the `control-label-group` element. This property value assumes that there are no
+     * special cases for which the label width differs. There is no way to get the width of the form label
+     * programmatically.
+     */
     margin-left: 14em;
     width: auto;
     justify-content: start;
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -1,0 +1,60 @@
+// Layout
+.callout {
+  display: flex;
+  column-gap: 1em;
+  justify-content: start;
+
+  &.callout-fit-content {
+    width: fit-content;
+  }
+
+  i.icon::before {
+    margin-right: 0;
+  }
+
+  p {
+    margin: 0;
+  }
+
+  .callout-title {
+    margin-bottom: .5em;
+  }
+
+  .callout-text {
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+// Style
+.callout {
+  padding: .5em 1em;
+  border: 1px solid var(--callout-color);
+  background-color: var(--callout-bg-color);
+  border-radius: .25em;
+
+  i.icon {
+    color: var(--callout-color);
+    font-size: 1.5em;
+  }
+
+  &.callout-type-info {
+    --callout-color: @callout-info-color;
+    --callout-bg-color: @callout-info-bg;
+  }
+
+  &.callout-type-success {
+    --callout-color: @callout-success-color;
+    --callout-bg-color: @callout-success-bg;
+  }
+
+  &.callout-type-warning {
+    --callout-color: @callout-warning-color;
+    --callout-bg-color: @callout-warning-bg;
+  }
+
+  &.callout-type-error {
+    --callout-color: @callout-error-color;
+    --callout-bg-color: @callout-error-bg;
+  }
+}
--- a/asset/css/variables.less
+++ b/asset/css/variables.less
@@ -130,6 +130,16 @@
 @schedule-element-fields-disabled-selected-bg: @base-gray-light;
 @schedule-element-keyboard-note-bg: @base-gray-light;
 
+@callout-success-color: @state-ok;
+@callout-info-color: #008fe8;
+@callout-warning-color: @state-warning;
+@callout-error-color: @state-critical;
+
+@callout-success-bg: fade(@callout-success-color, 10%);
+@callout-info-bg: fade(@callout-info-color, 10%);
+@callout-warning-bg: fade(@callout-warning-color, 10%);
+@callout-error-bg: fade(@callout-error-color, 10%);
+
 @empty-state-color: @base-gray-semilight;
 @empty-state-bar-bg: @base-gray-lighter;
 
--- a/src/Common/CalloutType.php
+++ b/src/Common/CalloutType.php
@@ -9,10 +9,10 @@ use ipl\Web\Widget\Icon;
  */
 enum CalloutType: string
 {
-    case Info = "info";
-    case Success = "success";
-    case Warning = "warning";
-    case Error = "error";
+    case Info = 'callout-type-info';
+    case Success = 'callout-type-success';
+    case Warning = 'callout-type-warning';
+    case Error = 'callout-type-error';
 
     /**
      * Get the icon element for use in the callout
@@ -21,11 +21,11 @@ enum CalloutType: string
      */
     public function getIcon(): Icon
     {
-        return match ($this) {
-            self::Info => new Icon('circle-info'),
-            self::Success => new Icon('circle-check'),
-            self::Warning => new Icon('warning'),
-            self::Error => new Icon('circle-xmark'),
-        };
+        return new Icon(match ($this) {
+            self::Info => 'circle-info',
+            self::Success => 'circle-check',
+            self::Warning => 'warning',
+            self::Error => 'circle-xmark',
+        });
     }
 }
--- a/src/Common/CalloutType.php
+++ b/src/Common/CalloutType.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use ipl\Web\Widget\Icon;
+
+/**
+ * An enum containing all possible callout types for the {@see Callout} widget.
+ */
+enum CalloutType: string
+{
+    case Info = "info";
+    case Success = "success";
+    case Warning = "warning";
+    case Error = "error";
+
+    /**
+     * Get the icon element for use in the callout
+     *
+     * @return Icon
+     */
+    public function getIcon(): Icon
+    {
+        return match ($this) {
+            self::Info => new Icon('circle-info'),
+            self::Success => new Icon('circle-check'),
+            self::Warning => new Icon('warning'),
+            self::Error => new Icon('circle-xmark'),
+        };
+    }
+}
--- a/src/Common/CalloutType.php
+++ b/src/Common/CalloutType.php
@@ -5,7 +5,7 @@ namespace ipl\Web\Common;
 use ipl\Web\Widget\Icon;
 
 /**
- * An enum containing all possible callout types for the {@see Callout} widget.
+ * An enum containing all possible callout types for the {@see Callout} widget
  */
 enum CalloutType: string
 {
--- a/src/Common/CalloutType.php
+++ b/src/Common/CalloutType.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use ipl\Web\Widget\Icon;
+
+/**
+ * An enum containing all possible callout types for the {@see \ipl\Web\Widget\Callout} widget
+ */
+enum CalloutType: string
+{
+    case Info = 'callout-type-info';
+    case Success = 'callout-type-success';
+    case Warning = 'callout-type-warning';
+    case Error = 'callout-type-error';
+
+    /**
+     * Get the icon element for use in the callout
+     *
+     * @return Icon
+     */
+    public function getIcon(): Icon
+    {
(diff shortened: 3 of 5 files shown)


flourish86 (dismissed):
lgtm

src/Common/CalloutType.php
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * An enum containing all possible callout types for the {@see Callout} widget.
+ */
+enum CalloutType: string
+{
+    case Info = "info";
+    case Success = "success";
+    case Warning = "warning";
+    case Error = "error";
```
Al2Klimov: Why the double quotes?

src/Widget/Callout.php
```diff
@@ -0,0 +1,104 @@
...
+     * @param ValidHtml|string $content the content of the callout
+     * @param string|null $title an optional title, displayed above the content
+     */
+    public function __construct(
+        protected CalloutType $type,
+        ValidHtml|string $content,
+        protected ?string $title = null
+    ) {
+        if (is_string($content)) {
+            $content = new Text($content);
```
Al2Klimov: With the above suggestion, you'd move this `new Text` into `assemble()`.

Al2Klimov (dismissed):
Tested together with https://github.com/Icinga/icingaweb2/pull/5477#pullrequestreview-4031546304.

jrauh01:
Would love to see this feature soon, so I don't have to create an own callout widget for https://github.com/Icinga/icingaweb2/pull/5430.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function __construct(
+        protected CalloutType $type,
+        protected ValidHtml|string $content,
+        protected ?string $title = null
+    ) {
+        $this->addAttributes(Attributes::create(['class' => $type->value]));
+    }
+
+    public function assemble(): void
```
jrauh01: Should be protected.

```suggestion
    protected function assemble(): void

File: src/Widget/Callout.php [hunk 4]
```diff
@@ -0,0 +1,95 @@
...
+     * @param bool $isFormElement should the callout be used for a form element
+     *
+     * @return $this
+     */
+    public function setFormElement(bool $isFormElement = true): static
+    {
+        if ($isFormElement) {
+            $this->addAttributes(Attributes::create(['class' => 'form-callout']));
+        } else {
+            $this->removeAttribute('class', 'from-callout');
```
jrauh01:
```suggestion
            $this->removeAttribute('class', 'form-callout');

File: src/Widget/Callout.php [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+use ipl\I18n\Translation;
+use ipl\Web\Common\CalloutType;
+
+/**
+ * An information box that can be used to display information to the user.
+ * It consists of a set of standardized colors and icons that can be used to visually distinguish the type of
+ * information.
+ * A content string and an optional title can be passed to the constructor.
+ */
+class Callout extends BaseHtmlElement
```
jrauh01, line 18: The formatting looks a bit odd. Please adjust this so that the line lengths are more equal. The line does not need to end after a full stop.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Callouts are only as wide as their content.
+     * Setting it to fullwidth will force the callout to be as wide as its container.
+     *
+     * @param bool $fullwidth should the callout be fullwidth
+     *
+     * @return $this
+     */
+    public function setFullwidth(bool $fullwidth = true): static
```
jrauh01: Maybe add a short description without a full stop in the first line of the docs block.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Setting this to true will allow the callout to be used for a single form element.
+     * This is used to visually align the callout to the content of the form element.
+     *
+     * @param bool $isFormElement should the callout be used for a form element
+     *
+     * @return $this
+     */
+    public function setFormElement(bool $isFormElement = true): static
```
jrauh01: Maybe add a short description without a full stop in the first line of the docs block.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        if ($isFormElement) {
+            $this->addAttributes(Attributes::create(['class' => 'form-callout']));
```
jrauh01: All CSS classes use the `callout-*` pattern, while this one is `*-callout`. I understand the intention to indicate that it’s a different kind of callout. Maybe it could be named something like `callout-form-element`. Another option would be to rename `callout-fullwidth` to `fullwidth-callout` to keep things consistent.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+    protected $tag = 'div';
+
+    protected $defaultAttributes = ['class' => 'callout'];
+
+    /**
+     * Create a new callout
+     *
+     * @param CalloutType $type the type of the callout. The type determines the color and icon that is used.
+     * @param ValidHtml|string $content the content of the callout
+     * @param string|null $title an optional title, displayed above the content
```
jrauh01: The first letter of each param doc should be uppercase.

src/Widget/Callout.php
```diff
@@ -0,0 +1,95 @@
...
+                is_string($this->content) ? new Text($this->content) : $this->content,
+            ],
+        ));
+    }
+
+    /**
+     * Callouts are only as wide as their content.
+     * Setting it to fullwidth will force the callout to be as wide as its container.
+     *
+     * @param bool $fullwidth should the callout be fullwidth
```
jrauh01: First letter should be uppercase.

```suggestion
     * @param bool $fullwidth Should the callout be fullwidth

File: src/Widget/Callout.php [hunk 11]
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        return $this;
+    }
+
+    /**
+     * Setting this to true will allow the callout to be used for a single form element.
+     * This is used to visually align the callout to the content of the form element.
+     *
+     * @param bool $isFormElement should the callout be used for a form element
```
jrauh01: First letter should be uppercase.

```suggestion
     * @param bool $isFormElement Should the callout be used for a form element

**Review by jrauh01 — CHANGES_REQUESTED** [review]
I tested the different kinds and types and think it looks great. First, please rebase.

File: src/Widget/Callout.php (hunk 11 above)
Comment (jrauh01) [reply src/Widget/Callout.php]: Maybe this could be changed to

```suggestion
     * @param bool $isFormElement Whether the callout should be used for a form element

File: src/Widget/Callout.php (hunk 10 above)
Comment (jrauh01) [reply src/Widget/Callout.php]: Maybe this could be changed to

```suggestion
     * @param bool $fullwidth Whether the callout should be full width

File: src/Widget/Callout.php (hunk 5 above)
Comment (TheSyscall) [reply src/Widget/Callout.php:18]: Changed

File: src/Widget/Callout.php (hunk 6 above)
Comment (TheSyscall) [reply src/Widget/Callout.php]: Added short description

File: src/Widget/Callout.php (hunk 7 above)
Comment (TheSyscall) [reply src/Widget/Callout.php]: Added short description

File: src/Widget/Callout.php (hunk 8 above)
Comment (TheSyscall) [reply src/Widget/Callout.php]: Changed to `callout-full-width` and `callout-form-element`

File: src/Widget/Callout.php (hunk 9 above)
Comment (TheSyscall) [reply src/Widget/Callout.php]: Changed to uppercase

File: src/Widget/Callout.php [hunk 12]
```diff
@@ -0,0 +1,107 @@
...
+    protected $tag = 'div';
+
+    protected $defaultAttributes = ['class' => 'callout'];
+
+    /**
+     * Create a new callout
+     *
+     * The $type parameter determines the color and icon of the callout.
+     *
+     * @param CalloutType $type The type of the callout.
```
jrauh01: Please remove the full stop.

```suggestion
     * @param CalloutType $type The type of the callout

File: src/Widget/Callout.php [hunk 13]
```diff
@@ -0,0 +1,107 @@
...
+ * An information box that can be used to display information to the user.
+ * It consists of a set of standardized colors and icons that can be used to
+ * visually distinguish the type of information. A content string and an
+ * optional title can be passed to the constructor.
+ */
+class Callout extends BaseHtmlElement
+{
+    use Translation;
+
+    /** @var string Classname for form-element callouts */
```
jrauh01:
```suggestion
    /** @var string Class name for form element callouts */

File: src/Widget/Callout.php [hunk 14]
```diff
@@ -0,0 +1,107 @@
...
+    protected const CLASS_FORM_ELEMENT = 'callout-form-element';
+
+    /** @var string Classname for full-width callouts */
```
jrauh01:
```suggestion
    /** @var string Class name for full width callouts */

File: src/Widget/Callout.php [hunk 15]
```diff
@@ -0,0 +1,107 @@
...
+    protected $defaultAttributes = ['class' => 'callout'];
+
+    /**
+     * Create a new callout
+     *
+     * The $type parameter determines the color and icon of the callout.
+     *
+     * @param CalloutType $type The type of the callout.
+     * @param ValidHtml|string $content The content of the callout
+     * @param string|null $title An optional title, displayed above the content
```
jrauh01:
```suggestion
     * @param ?string $title An optional title, displayed above the content

**Review by jrauh01 — CHANGES_REQUESTED** [review]
Just some minor phpdocs adjustments.

File: src/Widget/Callout.php [hunk 16]
```diff
@@ -0,0 +1,107 @@
...
+    protected function assemble(): void
+    {
+        $this->addHtml($this->type->getIcon());
+
+        $this->addHtml(HtmlElement::create(
+            'div',
+            ['class' => 'callout-text'],
+            [
+                $this->title
+                    ? HtmlElement::create('strong', ['class' => 'callout-title'], new Text($this->title))
```
jrauh01: Please use `Text::create()` factory.

```suggestion
                    ? HtmlElement::create('strong', ['class' => 'callout-title'], Text::create($this->title))

File: src/Widget/Callout.php [hunk 17]
```diff
@@ -0,0 +1,107 @@
...
+                    : null,
+                is_string($this->content) ? new Text($this->content) : $this->content,
```
jrauh01: Please use `Text::create()` factory.

```suggestion
                is_string($this->content) ? Text::create($this->content) : $this->content,

**Review by jrauh01 — CHANGES_REQUESTED** [review]

File: src/Widget/Callout.php [hunk 18]
```diff
@@ -0,0 +1,107 @@
...
+     * Set the callout width to be 100% of its parent container
+     *
+     * Callouts are normally only as wide as their content.
+     * Setting it to fullwidth will force the callout to be as wide as its container.
+     *
+     * @param bool $isFullWidth Whether the callout should be full width
+     *
+     * @return $this
+     */
+    public function setFullwidth(bool $isFullWidth = true): static
```
jrauh01: Regarding to the other changes, please adjust to camel case.

```suggestion
    public function setFullWidth(bool $isFullWidth = true): static

File: asset/css/callout.less [hunk 19]
```diff
@@ -0,0 +1,66 @@
...
+  width: fit-content;
+  margin: 0 auto 1em;
+
+  &.callout-full-width {
+    width: 100%;
+    justify-content: start;
+  }
+
+  &.callout-form-element {
+    margin-left: 14em;
```
jrauh01: Maybe this is worth an explaining comment.

jrauh01 (changes requested):

asset/css/callout.less
TheSyscall: Added a comment.

src/Common/CalloutType.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use ipl\Web\Widget\Icon;
+
+/**
+ * An enum containing all possible callout types for the {@see Callout} widget.
```
jrauh01: Please remove the period from the summary line.

```suggestion
 * An enum containing all possible callout types for the {@see Callout} widget

File: src/Widget/Callout.php [hunk 21]
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * An information box that can be used to display information to the user.
+ * It consists of a set of standardized colors and icons that can be used to
+ * visually distinguish the type of information. A content string and an
+ * optional title can be passed to the constructor.
+ */
+class Callout extends BaseHtmlElement
+{
+    use Translation;
+
```
jrauh01: This is never used and can be removed.

```suggestion

File: src/Widget/Callout.php [hunk 22]
```diff
@@ -0,0 +1,107 @@
...
+                is_string($this->content) ? Text::create($this->content) : $this->content,
+            ],
+        ));
+    }
+
+    /**
+     * Set the callout width to be 100% of its parent container
+     *
+     * Callouts are normally only as wide as their content.
+     * Setting it to fullwidth will force the callout to be as wide as its container.
```
jrauh01:
```suggestion      
     * Setting it to full width will force the callout to be as wide as its container.

File: src/Widget/Callout.php [hunk 23]
```diff
@@ -0,0 +1,107 @@
...
+                $this->title
+                    ? HtmlElement::create('strong', ['class' => 'callout-title'], Text::create($this->title))
+                    : null,
+                is_string($this->content) ? Text::create($this->content) : $this->content,
+            ],
+        ));
+    }
+
+    /**
+     * Set the callout width to be 100% of its parent container
```
jrauh01:
```suggestion
     * Set the callout width to 100% of its parent container

File: src/Widget/Callout.php [hunk 24]
```diff
@@ -0,0 +1,107 @@
...
+use ipl\Html\ValidHtml;
+use ipl\I18n\Translation;
+use ipl\Web\Common\CalloutType;
+
+/**
+ * An information box that can be used to display information to the user.
+ * It consists of a set of standardized colors and icons that can be used to
+ * visually distinguish the type of information. A content string and an
+ * optional title can be passed to the constructor.
+ */
```
jrauh01, line 17: Maybe this could be something like:

```suggestion
/**
 * Information box with a type specific color and icon
 *
 * The type controls both the color scheme and the icon. An optional title
 * is displayed above the content.
 */

**Review by jrauh01 — CHANGES_REQUESTED** [review]

File: src/Widget/Callout.php (hunk 21 above)
Comment (TheSyscall) [reply src/Widget/Callout.php]: Changed and removed import

**Review by jrauh01 — COMMENTED** [review]
Consider adding unit tests. Otherwise this looks good now.

File: src/Widget/Callout.php [hunk 25]
```diff
@@ -0,0 +1,105 @@
+<?php
+
+namespace ipl\Web\Widget;
+
+use InvalidArgumentException;
```
jrauh01: Unused import, please remove.

```suggestion

**Review by jrauh01 — CHANGES_REQUESTED** [review]

**Review by jrauh01 — APPROVED** [review]
LGTM

File: src/Widget/Callout.php [hunk 26]
```diff
@@ -0,0 +1,104 @@
...
+
+    protected function assemble(): void
+    {
+        $this->addHtml($this->type->getIcon());
+
+        $this->addHtml(HtmlElement::create(
+            'div',
+            ['class' => 'callout-text'],
+            [
+                $this->title !== null
```
lippserd: An empty string title renders a `<strong class="callout-title"></strong>` with no visible content. Consider adding a `trim() !== ''` check as in https://github.com/Icinga/ipl-validator/blob/main/src/RegexMatchValidator.php#L36. Also add a test for this. You could also create a PR in ipl-stdlib which adds `Str::isEmpty()` for future use.

lippserd (changes requested):
According https://github.com/Icinga/ipl-web/pull/358#issuecomment-4080781608:

Can you add a note in callout.less that this is a known limitation and max-width is not used in purpose?

`setFullWidth(true)` and `setFormElement(true)` can both be applied to the same instance with no documented combined behavior. In the LESS file, both modifiers declare width at equal specificity, so whichever declaration appears last in the stylesheet wins. Is this intentional? Shouldn't this be a mode setter which sets one class exclusively?

src/Widget/Callout.php
TheSyscall: Implemented `Str::isEmpty()` in https://github.com/Icinga/ipl-stdlib/pull/77.

Tests don't pass because of this.

lippserd (commented):
Looks good. I will approve once you have cleaned up your commits.

lippserd (dismissed):
> Looks good. I will approve once you have cleaned up your commits.

Wait a moment :). Please don’t use Str::isEmpty() in this PR. We can do this later. Else, this PR would require an stdlib release and a composer.json raise.

asset/css/callout.less
nilmerg, line 1: The variable usage here is rather sub-optimal. For one, `@color-…` represents a dependency on Icinga Web. You cannot use such variables in the ipl. Instead, introduce e.g. `@callout-success-color`, initialize it with `@state-ok` and declare it inside variables.less.

Do the same for the background, having to deal with functions in a theme is a nightmare. Even more so if there's no way to customize it.

nilmerg (changes requested):

src/Widget/Callout.php
TheSyscall: Reverted

asset/css/callout.less
TheSyscall: Added less color variables for text and background.

asset/css/variables.less
```diff
@@ -130,6 +130,16 @@
 @schedule-element-fields-disabled-selected-bg: @base-gray-light;
 @schedule-element-keyboard-note-bg: @base-gray-light;
 
+@callout-success-color: @state-ok;
+@callout-info-color: @base-primary-color;
```
nilmerg: I don't think *icinga blue* is what we should use here. You previously referenced the color for pending, which wasn't perfectly suitable either, but at least better than this. *icinga blue* is a brand color, which we already changed in the past, also it is used primarily as eye catcher. This is is for a box that informs the user and hence should default to a normal blue color, feel free to define one here for the first time if there's no matching base or default.

nilmerg (changes requested):

asset/css/callout.less
```diff
@@ -0,0 +1,61 @@
+// Layout
+.callout {
+  display: flex;
+  column-gap: 1em;
+  width: 100%;
```
nilmerg: This is bad practice, quite frankly. There's no need to force a width of 100%, it's a block element after all. It should expand depending on the parent as is normal for such.

nilmerg (changes requested):

asset/css/variables.less
TheSyscall: Changed to a blue info color

asset/css/callout.less
TheSyscall: Removed forced 100% width

nilmerg (approved):
LGTM :)

PR #361, opened 2026-03-23
Implement Csp

TheSyscall:
Add a Csp class which represents the Content-Security-Policy header.

This new class can be used to manage the content security policy.
This class enforces `default-src 'self'` as a baseline for a secure web application.
The first nonce added to any directive is selected as THE nonce for the CSP.
Setting a nonce is not supported since a nonce could appear in any number of different directives.


src/Common/Csp.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Get the values of a directive
+     *
+     * @param string $directive The directive name
+     *
+     * @return Generator<string>
+     */
+    public function getDirective(string $directive): Generator
+    {
+        yield from $this->directives[$directive] ?? [];
```
Al2Klimov: I wonder why you don't just return the array...

TheSyscall: Changed to return the array

Al2Klimov (changes requested):
(Meanwhile I'll review all together...

* https://github.com/Icinga/icingaweb2/pull/5477
* #358
* #361

src/Common/Csp.php
```diff
@@ -0,0 +1,175 @@
...
+
+namespace ipl\Web\Common;
+
+use InvalidArgumentException;
+
+/**
+ * Represents a Content Security Policy (CSP) header.
+ * Methods are additive, and duplicate policies are ignored.
+ */
+class Csp
```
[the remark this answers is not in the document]
TheSyscall: Added some tests

Al2Klimov (dismissed):
Tested together with https://github.com/Icinga/icingaweb2/pull/5477#pullrequestreview-4031546304.

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+        $header = str_replace("\n", ' ', $header);
+        $result = new static();
+        foreach (explode(';', $header) as $directive) {
+            $directive = trim($directive);
+            if (empty($directive)) {
+                continue;
+            }
+            $parts = explode(' ', $directive, 2);
+            if (count($parts) < 2) {
+                continue;
```
Al2Klimov: Consider throwing InvalidArgumentException in this method whenever you encounter invalid syntax. The already called add() already throws InvalidArgumentException. You decide.

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+        if ($policy === '*') {
+            return;
+        }
+
+        if (
+            (str_starts_with($policy, "'") && ! str_ends_with($policy, "'"))
+            || ! str_starts_with($policy, "'") && str_ends_with($policy, "'")
+        ) {
+            throw new InvalidArgumentException(
+                "Quoted policy must be fully surrounded by single quotes. policy: $policy",
```
Al2Klimov:
```suggestion
                "Quoted policy must be fully surrounded by single quotes. Policy: $policy",
```

Same below.

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        if (in_array('*', $policies)) {
+            return true;
+        }
+
+        $parsedUrl = parse_url($url);
+
+        if (! isset($parsedUrl['host'])) {
+            throw new InvalidArgumentException("URL must specify a host. url: $url");
```
Al2Klimov: If it's an invalid argument, why it's allowed above? ('none', *)

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        $scheme = $parsedUrl['scheme'] ?? null;
+        if (in_array("'self'", $policies)) {
+            $requestUri = ServerRequest::getUriFromGlobals();
+            if (
+                ($scheme === null || $requestUri->getScheme() === $scheme)
+                && $requestUri->getHost() === $parsedUrl['host']
```
Al2Klimov: Are you aware that && has a higher precedence than ||?

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+
+        $parsedUrl = parse_url($url);
+
+        if (! isset($parsedUrl['host'])) {
+            throw new InvalidArgumentException("URL must specify a host. url: $url");
+        }
+
+        $scheme = $parsedUrl['scheme'] ?? null;
+        if (in_array("'self'", $policies)) {
+            $requestUri = ServerRequest::getUriFromGlobals();
```
Al2Klimov: Please tell me evaluateUrl() isn't actually needed at all. 🙈

Because, due to reverse proxies in front of IW2, $_SERVER is so unreliable that https://github.com/Icinga/icinga-sso-web/pull/3 JS had to resolve the absolute URL on the client side...

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+            throw new InvalidArgumentException(
+                "Quoted policy must be fully surrounded by single quotes. policy: $policy",
+            );
+        }
+
+        if (str_starts_with($policy, "'") && str_ends_with($policy, "'")) {
+            return;
+        }
+
+        // scheme and scheme://*
```
Al2Klimov:
```suggestion
        // scheme: and scheme://*
```

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+                return true;
+            }
+        }
+
+        foreach ($policies as $policy) {
+            if (str_starts_with($policy, "'") && str_ends_with($policy, "'")) {
+                continue;
+            }
+
+            if ($scheme !== null && ($policy === $scheme . ':' || $policy === $scheme . '://')) {
```
Al2Klimov:
```suggestion
            if ($scheme !== null && ($policy === $scheme . ':' || $policy === $scheme . '://*')) {
```

(If I'm not mistaken.)

src/Common/Csp.php
```diff
@@ -0,0 +1,340 @@
...
+                && $parsedPolicyUrl['host'] === $parsedUrl['host']
+                && ($parsedPolicyPath === null || (
+                    $pathIsDirectory && $parsedPath !== null && str_starts_with($parsedPath, $parsedPolicyPath)
+                    || $parsedPath === $parsedPolicyPath
+                ))
+            ) {
+                return true;
+            }
+
+            // Note: https://*.example.com means https://example.com and https://sub.example.com
```
Al2Klimov: Then I'd explicitly check for the "."

src/Common/Csp.php
TheSyscall: It was implemented to support checking for user generated things navigation items and dashlets.
The idea was to give the users feedback on iframes that do not load or during the creation of new entries, specifically when using custom CSP.

src/Common/Csp.php
TheSyscall: It is not possible for anything else to be there.
`validatePolicy` already prohibits this.

src/Common/Csp.php
TheSyscall: Yes

src/Common/Csp.php
TheSyscall: Good argument. Changed

src/Common/Csp.php
TheSyscall: Yes. This allows us to write better exception messages.

src/Common/Csp.php
TheSyscall: Changed

src/Common/Csp.php
TheSyscall: Changed comment

src/Common/Csp.php
TheSyscall: You are not

src/Common/Csp.php
Al2Klimov: Famous last words before Claude Mythos roasts your codebase.

Seriously speaking:

* I see no validatePolicy() usage here
* Why not just to explicitly compare host === `*.`+host?

TheSyscall: - `validatePolicy` checks the policy during `add`. Validating all policies again seems unnecessary. The only way a policy could be added is by a subclass writing to the array directly.
- But yes, explicitly checking if the "." exists doesn't hurt. We can't do `*.`+host since `https://a.b.c.example.com` is also allowed by the expression `*.c.example.com`.

src/Common/Csp.php
Al2Klimov: But the function isn't used, even in Icinga/icingaweb2#5477.

TheSyscall: After talking with @lippserd we decided to move the issue of warning users about problematic iframes to a separate PR.

The idea is to actually load a small iframe in the browser.
Then we can detect if it actually loaded successfully and render an error with JavaScript.

So this method can be removed entirely.

https://github.com/Icinga/icingaweb2/pull/5477 will just show callouts with generic warnings until we implement the JS based solution.

src/Common/Csp.php
TheSyscall: This method has been removed. See: https://github.com/Icinga/ipl-web/pull/361#discussion_r3092936432

src/Common/Csp.php
```diff
@@ -0,0 +1,311 @@
...
+    /**
+     * Create a new CSP from a string
+     *
+     * @param string $header The CSP header string
+     *
+     * @return static A new CSP containing all directives from the input header
+     */
+    public static function fromString(string $header): static
+    {
+        $header = trim($header);
```
Al2Klimov:
```suggestion
```

You decide, but IMAO that's a no-op.

src/Common/Csp.php
```diff
@@ -0,0 +1,311 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use GuzzleHttp\Psr7\ServerRequest;
```
Al2Klimov:
```suggestion
```

You decide, but IMAO that's a no-op.

src/Common/Csp.php
```diff
@@ -0,0 +1,311 @@
...
+            if (! $this->canDirectiveBeEmpty($directive)) {
+                throw new InvalidArgumentException(
+                    "Directive $directive can not be empty."
+                );
+            }
+            if (! isset($this->directives[$directive])) {
+                $this->directives[$directive] = [];
+            }
+
+            if (in_array($value, $this->directives[$directive])) {
```
Al2Klimov: Will null ever be in the array?

src/Common/Csp.php
```diff
@@ -0,0 +1,311 @@
...
+                return $this;
+            }
+        } elseif (is_string($value)) {
+            $value = trim($value);
+
+            if (str_contains($value, ' ')) {
+                return $this->add($directive, explode(' ', $value));
+            }
+
+            if (empty($value)) {
```
Al2Klimov: ~Doesn't canDirectiveBeEmpty() matter?~ _Isn't validateExpression() enough?_

Al2Klimov: Here we go again... `php -r 'echo (int) empty("0");'`

src/Common/Csp.php
```diff
@@ -0,0 +1,311 @@
...
+            $this->validateExpression($value);
+
+            if (in_array($value, $this->directives[$directive] ?? [])) {
+                return $this;
+            }
+
+            if (! isset($this->directives[$directive])) {
+                $this->directives[$directive] = [];
+            }
+
```
Al2Klimov:
```suggestion
```

I wonder if this would work even with deprecation warnings enabled...

Al2Klimov (commented):
.

src/Common/Csp.php
TheSyscall: Why would this be an issue with deprecation warnings?
`isset` is made for exactly this job.

src/Common/Csp.php
TheSyscall: It is

TheSyscall: This is not the check I wanted to make. "0" is a valid string like any other.

Only numbers make little sense in the context of CSPs, ~but that is another issue and should be checked in `validateExpression`~ but technically any number is a valid http token that can be used in the `report-to` directive.

src/Common/Csp.php
TheSyscall: No. Changed it

src/Common/Csp.php
```diff
@@ -0,0 +1,497 @@
...
+            if (in_array($value, $this->directives[$directive] ?? [])) {
+                return $this;
+            }
+
+            if ($this->nonce === null && $this->isNonce($value)) {
+                $this->nonce = substr($value, 7, -1);
+            }
+
+            $this->directives[$directive] ??= [];
```
Al2Klimov:
```suggestion
```

Is this line necessary?

src/Common/Csp.php
```diff
@@ -0,0 +1,497 @@
...
+        }
+
+        // scheme: and scheme://*
+        if (preg_match('/^[a-z]+:(\/\/\*)?$/', $expression)) {
+            return;
+        }
+
+        preg_match(
+            '/^(?:(?<schema>[a-z]+):\/\/)?(?<host>(?:(?:[a-zA-Z0-9_\-\*]+|\*)\.)'
+            . '*(?:[\*a-z0-9]+))(?:\:(?<port>-?[0-9]+|\*))?(?<path>(?:\/[a-z0-9\%\-_\.]+)*)/',
```
Al2Klimov: Damn, if there only was a function to parse URLs...

src/Common/Csp.php
```diff
@@ -0,0 +1,497 @@
...
+        if ($parsedUrl['host'] === '') {
+            throw new InvalidArgumentException("Expression URL must specify a host. Expression: $expression");
+        }
+
+        if (str_starts_with($parsedUrl['host'], '*')) {
+            if (! str_starts_with($parsedUrl['host'], '*.')) {
+                throw new InvalidArgumentException("Wildcard host must be a full subdomain. Expression: $expression");
+            }
+        } elseif (str_contains($parsedUrl['host'], '*')) {
+            throw new InvalidArgumentException(
```
Al2Klimov, line 363: I hacked your code:

```
php > $csp->add('img-src','https://lolcat.*.example.com'); // Expected to throw
PHP Warning:  Uncaught InvalidArgumentException: Wildcards can only be used at the start of the host. Expression: https://lolcat.*.example.com in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php:331
Stack trace:
#0 /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php(203): ipl\Web\Common\Csp->validateExpression('img-src', 'https://lolcat....')
#1 php shell code(1): ipl\Web\Common\Csp->add('img-src', 'https://lolcat....')
#2 {main}
  thrown in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php on line 331

Warning: Uncaught InvalidArgumentException: Wildcards can only be used at the start of the host. Expression: https://lolcat.*.example.com in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php:331
Stack trace:
#0 /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php(203): ipl\Web\Common\Csp->validateExpression('img-src', 'https://lolcat....')
#1 php shell code(1): ipl\Web\Common\Csp->add('img-src', 'https://lolcat....')
#2 {main}
  thrown in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php on line 331
php > $csp->add('img-src','https://*.*.example.com'); // Expected to throw
php >
```

I didn't use AI, btw.

src/Common/Csp.php
```diff
@@ -0,0 +1,497 @@
...
+     *
+     * @return ?string[]
+     */
+    public function getRawDirective(string $directive): ?array
+    {
+        if (isset($this->directives[$directive])) {
+            return $this->directives[$directive];
+        }
+
+        return null;
```
Al2Klimov: Consider simplifying into one return via `??`.

tests/Common/CspTest.php
```diff
@@ -0,0 +1,1038 @@
...
+use ipl\Web\Common\Csp;
+use PHPUnit\Framework\Attributes\DataProvider;
+
+class CspTest extends TestCase
+{
+    public function testEmpty(): void
+    {
+        $csp = new Csp();
+
+        $this->assertInstanceOf(Csp::class, $csp);
```
Al2Klimov:
```suggestion
```

Sure, PHP is sometimes weird, but not THAT weird. Right... ?

Al2Klimov (commented):
It is fascinating how powerful browsers have become, yet a bit sad that now we have to protect web apps with barbwire and guard dogs...

src/Common/Csp.php
TheSyscall: Sadly the PHP parse_url function gets really confused when the defined URL doesn't come with a scheme and no port is explicitly defined.

Namely, the whole host is placed inside the path.

src/Common/Csp.php
TheSyscall: Turns out its wasn't

src/Common/Csp.php
TheSyscall: Multiple "*" inside the host part are never allowed. Added a check for that.

src/Common/Csp.php
TheSyscall: Simplified

tests/Common/CspTest.php
TheSyscall: That would be very weird. Removed.

Al2Klimov (approved):
This is just a lib and unit tests pass, so... 🤷‍♂️

src/Common/Csp.php
```diff
@@ -0,0 +1,499 @@
...
+                throw new InvalidArgumentException("Port must be a number. Expression: $expression");
+            }
+            if ($parsedUrl['port'] <= 0 || $parsedUrl['port'] > 0xFFFF) {
+                throw new InvalidArgumentException("Port must be between 0 and 65535. Expression: $expression");
+            }
+        }
+
+        if ($parsedUrl['path'] !== '') {
+            if (str_contains($parsedUrl['path'], '*')) {
+                throw new InvalidArgumentException("Wildcards can't be used in the path. Expression: $expression");
```
Al2Klimov:
```
php > $csp = new ipl\Web\Common\Csp;
php > $csp->add('img-src','https://example.com/path*'); // Expected to throw
php >
```

src/Common/Csp.php
```diff
@@ -0,0 +1,499 @@
...
+            }
+
+            if (! in_array($expression, static::AVAILABLE_KEYWORDS, true)) {
+                throw new InvalidArgumentException("Unsupported keyword. Expression: $expression");
+            }
+
+            return;
+        }
+
+        // scheme: and scheme://*
```
Al2Klimov, line 336:
```
php > $csp->add('script-src', 'https://'); // Expected to throw?
php > echo $csp;
script-src https://
php >
```

src/Common/Csp.php
```diff
@@ -0,0 +1,499 @@
...
+        if ($parsedUrl['port'] !== '') {
+            if ($parsedUrl['port'] === '*') {
+                return;
+            }
+
+            if (! is_numeric($parsedUrl['port'])) {
+                throw new InvalidArgumentException("Port must be a number. Expression: $expression");
+            }
+            if ($parsedUrl['port'] <= 0 || $parsedUrl['port'] > 0xFFFF) {
+                throw new InvalidArgumentException("Port must be between 0 and 65535. Expression: $expression");
```
Al2Klimov:
```
php > $csp = new ipl\Web\Common\Csp;
php > $csp->add('script-src', 'https://example.com:0'); // Exception contradicts itself
PHP Warning:  Uncaught InvalidArgumentException: Port must be between 0 and 65535. Expression: https://example.com:0 in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php:351
Stack trace:
#0 /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php(203): ipl\Web\Common\Csp->validateExpression('script-src', 'https://example...')
#1 php shell code(1): ipl\Web\Common\Csp->add('script-src', 'https://example...')
#2 {main}
  thrown in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php on line 351

Warning: Uncaught InvalidArgumentException: Port must be between 0 and 65535. Expression: https://example.com:0 in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php:351
Stack trace:
#0 /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php(203): ipl\Web\Common\Csp->validateExpression('script-src', 'https://example...')
#1 php shell code(1): ipl\Web\Common\Csp->add('script-src', 'https://example...')
#2 {main}
  thrown in /Users/aklimov/NET/WS/ipl-web/src/Common/Csp.php on line 351
php >
```

src/Common/Csp.php
```diff
@@ -0,0 +1,499 @@
...
+                if (strlen($expression) < 9) {
+                    throw new InvalidArgumentException("Nonce must have a value. Expression: $expression");
+                }
+
+                return;
+            }
+
+            if (str_starts_with($expression, "'report-") && $expression !== "'report-sample'") {
+                foreach (static::AVAILABLE_HASHES as $hash) {
+                    if (str_ends_with($expression, sprintf("-%s'", $hash))) {
```
Al2Klimov:
```
php > $csp = new ipl\Web\Common\Csp;
php > $csp->add('script-src', "'report-sha256-sha384'"); // AI says it should throw, I'm not sure though
php >
```

Al2Klimov (changes requested):
I'm sorry. I almost forgot I have AI:

src/Common/Csp.php
TheSyscall: It should throw. There are only a set number of allow report hash algorithms.
Fixed validation.

src/Common/Csp.php
TheSyscall: Changed port range in exception to 1 to 65535

src/Common/Csp.php
TheSyscall: Added test and validation

src/Common/Csp.php
TheSyscall: There is never a situation where the path should have a '*'. Added Test abd validation.

src/Common/Csp.php
```diff
@@ -0,0 +1,510 @@
...
+            }
+
+            if ($parsedUrl['port'] <= 0 || $parsedUrl['port'] > 0xFFFF) {
+                throw new InvalidArgumentException("Port must be between 1 and 65535. Expression: $expression");
+            }
+        }
+
+        if ($parsedUrl['path'] !== '') {
+            if (str_contains($parsedUrl['path'], '*')) {
+                throw new InvalidArgumentException("Wildcards can't be used in the path. Expression: $expression");
```
Al2Klimov:
```
php > require_once 'src/Common/Csp.php';
php > $csp = new ipl\Web\Common\Csp;
php > $csp->add('script-src', 'https://example.com:*/path*'); // no exception!
php > $csp->add('script-src', 'https://example.com:*/*');     // no exception!
php >
```

Al2Klimov (changes requested):
I explicitly asked the AI for your URL parsing regex, as I'm skeptical. It didn't find anything there, so one more thing and we're done:

src/Common/Csp.php
TheSyscall: Fixed by @lippserd

Al2Klimov (approved):

PR #366, opened 2026-04-10
Support rich labels in `SearchSuggestions`

### LINKED ISSUE #364: Support rich labels in `\ipl\Web\FormElement\SearchSuggestions`
`\ipl\Web\FormElement\SearchSuggestions` currently only allows string labels and defaults to `<input type="button" value="<label>">`.

A suggestion's label should also be of type `\ipl\Html\ValidHtml` and this should result in a `<button type="button"><valid-html></button>` instead.

BastianLedererIcinga:
Allow providers of `SearchSuggestions` to pass `ValidHtml` via the `details` key, which is rendered inside a `button` element.

resolve #364

### DIFF
--- a/asset/css/search-base.less
+++ b/asset/css/search-base.less
@@ -84,7 +84,7 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   }
 }
 
-.search-suggestions {
+.search-suggestions.search-suggestions {
   background: var(--suggestions-bg, @suggestions-bg);
   color: var(--suggestions-color, @suggestions-color);
   border: 1px solid var(--suggestions-border-color, @suggestions-border-color);
@@ -355,7 +355,7 @@ fieldset:disabled .term-input-area [data-drag-initiator] {
   }
 }
 
-.search-suggestions {
+.search-suggestions.search-suggestions {
   z-index: 2; // Required so that nothing else can overlap it (such as opaque elements and the impact overlay)
   position: absolute;
   overflow: auto;
--- a/src/FormElement/SearchSuggestions.php
+++ b/src/FormElement/SearchSuggestions.php
@@ -6,6 +6,7 @@ use ipl\Html\Attributes;
 use ipl\Html\BaseHtmlElement;
 use ipl\Html\HtmlElement;
 use ipl\Html\Text;
+use ipl\Html\ValidHtml;
 use ipl\I18n\Translation;
 use Psr\Http\Message\ServerRequestInterface;
 use Traversable;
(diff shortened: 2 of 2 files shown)


asset/css/search-base.less
nilmerg, line 1: I'm not a fan of raising the CSS specificity this way. Instead, this is a case when markup has to change since the current CSS rules are independent anyway and target an element that isn't bound to any parent.

The `TermInput` is responsible to place its suggestions correctly and it can do so by adding them outside of the containing form. Add this to `TermInput::onRegistered()`:

```php
        $suggestions = (new HtmlElement('div'))
            ->setAttribute('id', Attribute::sanitizeId($this->getValueOfNameAttribute()) . '-suggestions')
            ->setAttribute('class', 'search-suggestions');

        $form->prependWrapper(
            (new HtmlDocument())
                ->addHtml($form, $suggestions)
        );
```

src/FormElement/SearchSuggestions.php
```diff
@@ -251,18 +253,31 @@ protected function assemble(): void
                     'type' => 'button',
                     'value' => $data['label'] ?? $data['search']
                 ];
+                $labelHtml = $data['label-html'] ?? null;
+                unset($data['label-html']);
```
nilmerg: What about `details` as key? (Which still replace `label` as content.)

nilmerg (changes requested):

nilmerg (approved):
@raviks789 Please test this with your Director customvar changes, since you make heavy use of the `TermInput` there.

The location of `.suggestions` changed to outside the form now and accumulate there in case of multiple instances.

raviks789:
@nilmerg just did a simple test. The suggestions are not displayed for data lists with this branch. See the screen shots below:

main branch:
Screenshot 2026-04-14 at 16 17 30



feature/rich-suggestion-labels branch:
Screenshot 2026-04-14 at 16 11 14

BastianLedererIcinga:
I could not think of a clever solution that fixes the problem with the fieldsets, but maybe Claude did:
https://github.com/Icinga/ipl-html/compare/main...fieldset-on-register
It fixes the problem in the director as far as I tested it but I'm not sure if this is 100% reliable.

I left the comments it made for now so it's easier to evaluate if the approach makes sense at all, if we want to use it I'm going to clean it up.

nilmerg:
I'm not sure whether this approach is correct. It looks promising, but it changes the timing of when fieldsets are assembled. Though, at least only for fieldsets. But I'd like to perform proper testing and right now there's not enough time for this. So please open a separate pull request for it and restore the `.search-suggestions.search-suggestions` approach here.

nilmerg (changes requested):
See above.

nilmerg (approved):

PR #373, opened 2026-04-24
Fix animation of appended relative time elements

### LINKED ISSUE #372: `RelativeTimeBehavior` does not detect changes in appended content
That's due to the selector `.container` being used for the `rendered` event. For appended content in containers this might not be true. An example are the history lists in Icinga DB Web.

BastianLedererIcinga:
fixes: #372 

To ensure `scan()` is run on appended content, it is now called only when the event has reached the container that is closest the its origin.
While working on this I noticed that the `Set` of `WeakRefs` used by `RelativeTime` stacked up several `WeakRefs` pointing to the same element when a container with known elements was rescanned, so I added `WeakSet` which is used by `scan` to prevent duplicates.

As suggested in https://github.com/Icinga/icingadb-web/pull/1322#pullrequestreview-4168528222 hidden elements are no longer animated, though this causes elements that become visible again to show an outdated value until the next tick.

### DIFF
--- a/asset/js/RelativeTime.js
+++ b/asset/js/RelativeTime.js
@@ -46,6 +46,10 @@ define(function () {
         }
 
         updateElement(element) {
+            if (element.hidden) {
+                return;
+            }
+
             const relativeTimeAgo = this.getType(element);
             if (relativeTimeAgo === 'ago' || relativeTimeAgo === 'since') {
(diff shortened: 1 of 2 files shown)


nilmerg (approved):
Well done, thanks!

PR #374, opened 2026-04-27
`CopyToClipboard`: Preserve inline layout of visually hidden labels

fsalomon:
Switch from `Element.innerText` to a DOM walk that concatenates text node contents directly. `innerText` is rendering-aware and inserts a newline around any element it considers block-level, including `sr-only` content.

`<br>` is converted to `\n` and elements hidden via `display: none` /
`visibility: hidden` are skipped, so script/style content is not copied.

This is otherwise functionally identical to `Range.toString()` on `selectNodeContents` (see Icinga/icingadb-web#1366).


nilmerg (approved):
Thank you!

PR #383, opened 2026-05-21
Handle search suggestions errors

nilmerg:
`\ipl\Web\Control\SearchBar\Suggestions` now catch `\ipl\Web\Control\SearchBar\SearchException` during assembly. Also, an unused CSS class `has-more` is not added anymore, as the initially planned pagination was never implemented and the refactor now made it unworthy to keep.

`\ipl\Web\FormElement\SearchSuggestions`'s constructor argument is now optional and a setter to show a failure message is added. In addition to that, it catches `\Exception` during assembly and uses it to indicate a failure on its own.

### DIFF
--- a/src/Control/SearchBar/Suggestions.php
+++ b/src/Control/SearchBar/Suggestions.php
@@ -220,18 +220,13 @@ abstract class Suggestions extends BaseHtmlElement
         $this->prependHtml(new HtmlElement('li', Attributes::create(['class' => 'default']), $button));
     }
 
-    protected function assemble()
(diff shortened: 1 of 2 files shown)


BastianLedererIcinga (approved):
Looks good, I made some local modifications to icingadb-web's `QueryColumnsProvider` and my [notifications-web PR](https://github.com/Icinga/icinga-notifications-web/pull/469) to test the exception handling for both suggestion types.
I will adjust the latter to rely on these changes.

PR #384, opened 2026-05-26
Show no placeholders if a `SearchEditor` has no suggestions

BastianLedererIcinga:
The placeholders of a `SearchEditor` currently always state `Type to search...`, which is rather misleading when an editor is intentionally created without suggestions. 

The `SearchEditor` is adjusted so the placeholders are only shown when suggestions are actually available.

### DIFF
--- a/src/Control/SearchEditor.php
+++ b/src/Control/SearchEditor.php
@@ -493,7 +493,9 @@ class SearchEditor extends Form
             'data-type' => 'column',
             'data-enrichment-type' => 'completion',
             'data-term-suggestions' => '#search-editor-suggestions',
-            'placeholder' => $this->translate('Start typing to search for a column')
+            'placeholder' => $this->getSuggestionUrl() !== null
+                ? $this->translate('Start typing to search for a column')
+                : null
         ]);
         $columnInput->getAttributes()->registerAttributeCallback('data-suggest-url', function () {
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
Now that I saw which text you want to show instead, I think showing nothing is the better alternative and this proposal should be changed in order to only show the default placeholder if a suggestion url is set.

Your alternative proposals…

* Enter a column
* Enter a value

…add nothing useful beyond the information added to tell the user no validation is happening. Let's focus on that to make it useful and really helpful instead of repeating such a generic placeholder like these over and over.

src/Control/SearchEditor.php
```diff
@@ -493,7 +493,9 @@ protected function createCondition(Filter\Condition $condition, $identifier)
             'data-type' => 'column',
             'data-enrichment-type' => 'completion',
             'data-term-suggestions' => '#search-editor-suggestions',
-            'placeholder' => $this->translate('Start typing to search for a column')
+            'placeholder' => $this->getSuggestionUrl() !== null
+                ? $this->translate('Start typing to search for a column')
+                : ''
```
nilmerg: please fall back to null instead of the empty string, to make sure the attribute isn't rendered.

nilmerg (approved):
Now please also update the title of the PR and we're done :P

PR #385, opened 2026-05-27
Introduce `DisplayFormElement` for embedding HTML in `CompatForm`

jrauh01:
Allows arbitrary `ValidHtml` content to be placed inside a `CompatForm` control group, complete with an optional label and description icon. The companion CSS rule ensures the wrapper div stretches to fill available horizontal space within the flex layout.

This is used for the new two-factor auth feature in Icinga Web.

### DIFF
--- a/src/Compat/DisplayFormElement.php
+++ b/src/Compat/DisplayFormElement.php
@@ -54,7 +54,7 @@ class DisplayFormElement extends BaseHtmlElement
             HtmlElement::create(
                 'div',
                 Attributes::create(['class' => 'control-label-group']),
-                $this->label ? HtmlElement::create('label', null, Text::create($this->label)) : null
+                $this->label !== null ? HtmlElement::create('label', null, Text::create($this->label)) : null
             )
         );
         $this->addHtml(
--- a/src/Compat/DisplayFormElement.php
+++ b/src/Compat/DisplayFormElement.php
@@ -7,6 +7,7 @@ namespace ipl\Web\Compat;
 
 use ipl\Html\Attributes;
 use ipl\Html\BaseHtmlElement;
+use ipl\Html\Contract\HtmlElementInterface;
 use ipl\Html\HtmlElement;
 use ipl\Html\Text;
 use ipl\Html\ValidHtml;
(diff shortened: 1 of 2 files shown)


src/Compat/DisplayFormElement.php
```diff
@@ -0,0 +1,69 @@
...
+            HtmlElement::create(
+                'div',
+                Attributes::create(['class' => 'control-label-group']),
+                $this->label ? HtmlElement::create('label', null, Text::create($this->label)) : null
+            )
+        );
+        $this->addHtml(
+            HtmlElement::create('div', Attributes::create(['class' => 'display-form-element']), $this->content)
+        );
+        if ($this->description) {
```
TheSyscall: Passing '0' to the description field behaves differently than in `DescriptionDecorator`.

Consider checking against null explicitly or use https://github.com/Icinga/ipl-stdlib/pull/77

src/Compat/DisplayFormElement.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->addHtml(
+            HtmlElement::create(
+                'div',
+                Attributes::create(['class' => 'control-label-group']),
+                $this->label ? HtmlElement::create('label', null, Text::create($this->label)) : null
+            )
+        );
+        $this->addHtml(
+            HtmlElement::create('div', Attributes::create(['class' => 'display-form-element']), $this->content)
```
TheSyscall, line 61: Consider adding `aria-describedby` like `DescriptionDecorator` does

src/Compat/DisplayFormElement.php
```diff
@@ -0,0 +1,69 @@
...
+        );
+        if ($this->description) {
+            $this->addHtml(
+                new Icon('info-circle', Attributes::create(['class' => 'control-info', 'title' => $this->description]))
```
TheSyscall, line 65: Consider adding `aria-hidden` and `role` like `DescriptionDecorator`

TheSyscall (changes requested):

src/Compat/DisplayFormElement.php
jrauh01: Done

src/Compat/DisplayFormElement.php
jrauh01: Done

src/Compat/DisplayFormElement.php
jrauh01: Done

src/Compat/DisplayFormElement.php
```diff
@@ -0,0 +1,69 @@
...
+    ) {
+    }
+
+    protected function assemble(): void
+    {
+        $this->addHtml(
+            HtmlElement::create(
+                'div',
+                Attributes::create(['class' => 'control-label-group']),
+                $this->label ? HtmlElement::create('label', null, Text::create($this->label)) : null
```
TheSyscall:
```suggestion
                $this->label !== null ? HtmlElement::create('label', null, Text::create($this->label)) : null
```

TheSyscall (changes requested):

TheSyscall (approved):
LGTM 👍

PR #389, opened 2026-06-15
PrimaryButtonDecorator: Only mark a single button primary

nilmerg:
It applies now only to implementations of `\ipl\Html\Form`. But the reality was that it previoulsy relied on `\ipl\Html\Contract\FormElements` instead which the method's interface didn't guarantee either.

closes #388

### DIFF
--- a/src/Compat/FormDecorator/PrimaryButtonDecorator.php
+++ b/src/Compat/FormDecorator/PrimaryButtonDecorator.php
@@ -6,16 +6,21 @@ use ipl\Html\Attribute;
 use ipl\Html\Contract\DecorationResult;
 use ipl\Html\Contract\Form;
 use ipl\Html\Contract\FormDecoration;
-use ipl\Html\Contract\FormSubmitElement;
+use ipl\Html\Form as iplForm;
 
+/**
+ * Marks the form’s submit button as primary (adds the "btn-primary" CSS class).
+ *
+ * Note: This decorator only applies to {@see \ipl\Html\Form} instances because it relies on
+ * the submit-button tracking provided by that implementation.
+ */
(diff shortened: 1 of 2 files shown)


src/Compat/FormDecorator/PrimaryButtonDecorator.php
```diff
@@ -6,16 +6,18 @@
...
-            if ($element instanceof FormSubmitElement) {
-                $element->getAttributes()->addAttribute(new Attribute('class', 'btn-primary'));
-            }
+        $submitButton = null;
+        if ($form instanceof iplForm && $form->hasSubmitButton()) {
+            $submitButton = $form->getSubmitButton();
         }
+
+        $submitButton?->getAttributes()
+            ->addAttribute(new Attribute('class', 'btn-primary'));
```
sukhwinder33445: Is there a reason why you didn't add the attribute directly in the condition above?

sukhwinder33445: By making only the first button primary, the styles of the other buttons change slightly — for example, the delete button looks bit different now. However, it still looks good and might actually be a fix rather than a problem. Before this decorator was added, the delete button was already styled like the after case shown below:

## Before
Screenshot 2026-06-15 at 19 53 24

## After
Screenshot 2026-06-15 at 19 53 53

(Screenshots from Contact group form of Icinga Notifications Web)

nilmerg: > Is there a reason why you didn't add the attribute directly in the condition above?

Not really 😅

sukhwinder33445 (approved):

PR #390, opened 2026-06-17
Fix long error messages in the `SearchEditor`

BastianLedererIcinga:
fix #382 
Ensure long validation messages span across the entire condition instead of pushing elements away.

Screenshot from 2026-06-24 08-50-13

### DIFF
--- a/asset/css/search-editor.less
+++ b/asset/css/search-editor.less
@@ -59,16 +59,14 @@
     .rounded-corners(0 .4em .4em 0);
   }
 
-  .search-error {
-    input:invalid {
-      background: var(--search-term-invalid-bg, @search-term-invalid-bg);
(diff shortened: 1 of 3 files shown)


src/Compat/FormDecorator/ConditionErrorsDecorator.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace ipl\Web\Compat\FormDecorator;
```
nilmerg: In it's current form it is rather limited in functionality and makes only sense combined with fieldsets that are set up how the search editor does. So please put this into a different (new) namespace: `ipl\Web\Control\SearchEditor`

nilmerg (changes requested):

nilmerg (approved):

PR #391, opened 2026-06-29
CsrfCounterMeasure: skip Sec-Fetch-Site check for safe HTTP methods

TheSyscall:
[RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#section-9.2.1) defines GET, HEAD, OPTIONS, and TRACE as safe methods: they must not cause side effects, so they cannot be exploited for CSRF. Previously, `requestIsSafe()` checked the Sec-Fetch-Site header unconditionally, which caused it to return false (and throw "Rejecting cross-site request") when a cross-origin GET navigation reached a form. This broke legitimate flows where an external service redirects the browser to a form. The redirect carries
`Sec-Fetch-Site: cross-site`, and the form assembly failed before the page could render.

For safe methods, `requestIsSafe()` now returns null, falling through to the token-based CSRF fallback. The token embedded in the rendered form still protects the subsequent POST submission.

### DIFF
--- a/src/Common/CsrfCounterMeasure.php
+++ b/src/Common/CsrfCounterMeasure.php
@@ -134,13 +134,21 @@ trait CsrfCounterMeasure
      * Get whether the request is safe from CSRF based on the Sec-Fetch-Site header
      *
      * Returns true if the header indicates a same-origin request or a user-originated operation (e.g. typing a URL),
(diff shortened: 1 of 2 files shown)


Al2Klimov (approved):
# 🙈

PR #392, opened 2026-06-29
Add `ViewModeSwitcher` control

### LINKED ISSUE #340: Introduce `ipl\Web\Control\ViewModeSwitcher`
Icinga DB Web provides its own control to let users choose between a list view's layout since its initial release. Since then, there were multiple attempts to generalize this, but only the last is, how I see it, final. See https://github.com/Icinga/ipl-web/pull/252 and related subsequent changes.

The widget to *control* this however, is still part of Icinga DB Web. We should transfer this here as well now.

#### Compatibility

There's not much here, only the way how it's prepared. Icinga DB Web provides `\Icinga\Module\Icingadb\Web\Controller::createViewModeSwitcher()` for this. Adding the same method to `CompatController`, like for the other controls (e.g. `createSortControl()`), is not possible as this would break older versions of Icinga DB Web due to the different method signature. I'd rather suggest to introduce a new trait `ipl\Web\Common\Controls` that provides a way to prepare the new view mode switcher. It may even be what provides the other preparation methods in the future, as requiring to extend `CompatController` was already not always an option, hence the generalized name *Controls*.

#### Extendability

We can find two cases where the default view modes were not sufficient. It can safely be assumed that this might be the case in the future again, in a different product. Icinga DB Web currently supports a fourth view mode: `tabular`. This is specific to Icinga DB Web and cannot supported by the new mode switcher. Then there is the `GridViewModeSwitcher` which overrides the available modes entirely to introduce its own. Both are valid use cases, though. So the new implementation should allow to add new modes as well as to remove default modes.

#### Integration

That's were things get difficult. These areas must be considered and still be supported.

##### Accessibility

The control is implemented as a group of radio elements which rely on unique IDs. With Icinga Web, these must be based on the container id as usual. This must still be possible.

##### Interaction With Other Controls

Icinga DB Web's view mode switcher cannot be prepared without a limit control. The reason is that changing to the minimal layout, the default limit should be 50 instead of 25. This might still be supported, but shouldn't be a requirement. The current implementation also supports Icinga DB Web's custom `grid` mode, so there's obviously the need for custom rules. I don't have a specific implementation in mind right now, but since the `LimitControl` is a part of ipl web already, it is certainly not a bad idea to try to preserve this. Not sure though, if we can or should support the case for a *vertical pagination*, maybe that's something only a custom rule is able to support.

##### Persistence

Changing the view mode in Icinga DB Web is persisted into the session/preferences of the user, dependent on where the change happened. That is, if you switch to the minimal mode in the host list, the service list still shows the default view mode, every host list though, whether the host overview, the host problems or a hosts dashlet (without an explicit view mode), all are shown in the minimal mode by default now. This is definitely not something the library can provide by default, but shouldn't make it hard to implement either. Custom events may be the answer to this. Though, at the moment this relies on form success handling, which in turn relies on when the http request is being handled. I'd like to avoid a solution like with `\ipl\Web\Compat\SearchControls::callHandleRequest()` at all cost!

So please let's think about an alternative. Working with promises in the past (mainly with JS), I'd imagine something like a promise that may be already settled and the handler is triggered immediately once registered. This is different to our current event implementation, as once the event has already been triggered, no handler registered later will be called anymore. This may even prove beneficial for other cases? Though, this is just an idea but I don't know if it's good or bad. Nor how it should be implemented and where. Whoever takes this on, should discuss it with @lippserd first.

But it may also be just a question of orchestration. I've already mentioned that the trait `ipl\Web\Common\Controls` may be home to the other preparation methods in the future. And the problem if and when the request is being handled isn't unique to the search bar nor to the new view mode switcher. So the answer to this may also be just a method `ipl\Web\Common\Controls::handleControls(ServerRequestInterface): static` which tracks prepared controls on its own and allows the controller to decide when the request is handled. The neat effect of course would be that custom event handlers may easily be registered beforehand.

We should find a way that's compatible with all controls right now, not just the view mode switcher. The above are only rough ideas and I didn't check their implication with each control. If the chosen idea/solution does not support all current controls, it's bad.

### Implementation

I don't really want to specify the actual implementation due to the considerations above. There's just too many of them to do this beforehand. I just want to point out, that this task requires thorough preparation and consideration of separation of concerns (Soc) is key.

My basic assumption is that there's:

* `ipl\Web\Control\ViewModeSwitcher`: A copy of `\Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher`, more or less
* `ipl\Web\Common\Controls`: A trait providing `createViewModeSwitcher(…)`. I've left out the method signature on purpose, as this is something that cannot be specified right now

BastianLedererIcinga:
resolve #340 

A `ViewModeSwitcher` class similar to the one in icingadb-web is added. It was reworked so that adding and removing view modes can be done without declaring a subclass.
To allow this the `ViewMode` class is added, which contains name, icon, titles for active/inactive state and the default page size.

The `Controls` trait is added with a `createViewModeSwitcher()` function.
It registers a listener to `ViewModeSwitcher::ON_REQUEST`, which populates the `ViewModeSwitcher` with the view mode from the url and emits `ON_VIEW_MODE_SET`.
If the view mode is changed, a `ViewModeSwitcher::ON_SUBMIT` listener emits `ON_VIEW_MODE_CHANGE` with the `ViewModeSwitcher`, its previous view mode and a `$redirectUrl` that listeners may modify.

The `handleControls()` function suggested in #340 is implemented, which calls `handleRequest()` for all controls passed to `trackControl()` , in case of the `ViewModeSwitcher` this is done by the factory method.

The `CompatController` registers listeners on the trait's events in the `createLimitControl()` and `createPaginationControl()` factories, to adjust the default limit, page size to teh current view mode, and correct the current page when the view mode is changed.

`SearchControls::createSearchBar()` is adjusted to shift the view mode param before building its filter, because the `Controls` trait does not have access to the `UrlParams`.

### DIFF
--- a/asset/css/controls.less
+++ b/asset/css/controls.less
@@ -109,10 +109,10 @@
 /**
  The default layout of list controls in Icinga Web
 
- ┌────────────────────────────────────────────────────────────────┐
- │ .pagination-control               .limit-control .sort-control │
- │ <-------------------- .search-controls ----------------------> │
- └────────────────────────────────────────────────────────────────┘
+ ┌───────────────────────────────────────────────────────────────────────────────────┐
+ │ .pagination-control              .view-mode-switcher .limit-control .sort-control │
+ │ <---------------------------- .search-controls ---------------------------------> │
+ └───────────────────────────────────────────────────────────────────────────────────┘
  */
 .controls.default-layout {
   .box-shadow(0, 0, 0, 1px, @controls-separator-bg);
@@ -127,10 +127,12 @@
   }
 
   > .sort-control,
+  > .view-mode-switcher,
   > .limit-control {
     float: right;
   }
 
+  > .view-mode-switcher,
   > .limit-control {
     margin-right: .5em;
   }
--- a/asset/css/view-mode-switcher.less
+++ b/asset/css/view-mode-switcher.less
@@ -1,0 +1,42 @@
+.view-mode-switcher {
+  display: flex;
+
+  input {
+    display: none;
+  }
+
+  label {
+    color: var(--control-color, @control-color);
+    line-height: 1;
+    background: var(--default-input-bg, @default-input-bg);
+    padding: 14/16*.25em 14/16*.5em;
+    font-size: 16/12em;
+    height: 24/16em; // desired pixel height / font-size
+    cursor: pointer;
+
+    &:first-of-type {
+      border-top-left-radius: 0.25em;
+      border-bottom-left-radius: 0.25em;
+    }
+
+    &:last-of-type {
+      border-top-right-radius: 0.25em;
+      border-bottom-right-radius: 0.25em;
+    }
+
+    &:not(:last-of-type) {
+      border-right: 1px solid var(--default-input-hover-bg, @default-input-hover-bg);
+    }
+
+    i {
+      // fix height for Chrome
+      display: block;
+    }
+  }
+
+  input[checked] + label {
+    background-color: var(--control-color, @control-color);
+    color: var(--default-text-color-inverted, @default-text-color-inverted);
+    cursor: default;
+  }
+}
--- a/src/Common/Controls.php
+++ b/src/Common/Controls.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use ipl\Html\Form;
+use ipl\Stdlib\Events;
+use ipl\Web\Control\ViewModeSwitcher;
+use ipl\Web\Url;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Provides factory methods to prepare reusable web controls and allows to handle their requests
+ *
+ * {@see static::handleControls()} can be called to call {@see Form::handleRequest()} for all Controls created by
+ * this trait.
+ */
+trait Controls
+{
+    use Events;
+
+    /** @var string Event emitted when the view mode has been changed */
+    public const ON_VIEW_MODE_CHANGE = 'view-mode-change';
+
+    /** @var string Event emmitted when the created {@see ViewModeSwitcher} is populated with a view mode */
+    public const ON_VIEW_MODE_SET = 'view-mode-set';
+
+    /** @var array<Form> Controls for which {@see self::handleControls()} must call {@see Form::handleRequest()} */
+    private array $trackedControls = [];
+
+    /**
+     * Redirect to the given Url
+     *
+     * @param Url|string $url
+     *
+     * @return never
+     */
+    abstract protected function redirectNow($url);
+
+    /**
+     * Create a {@see ViewModeSwitcher} control
+     *
+     * On {@see ViewModeSwitcher::ON_REQUEST} the created instance is populated with the view mode from the url
+     * and {@see static::ON_VIEW_MODE_SET} is emmitted.
+     *
+     * On {@see ViewModeSwitcher::ON_SUBMIT}, if the view mode was changed, {@see static::ON_VIEW_MODE_CHANGE}
+     * is emmitted with the {@see ViewModeSwitcher}, its previous view mode and a redirect url, which listeners
+     * may modify.
+     *
+     * @return ViewModeSwitcher
+     */
+    public function createViewModeSwitcher(): ViewModeSwitcher
+    {
+        $viewModeSwitcher = new ViewModeSwitcher();
+
+        $this->trackControl($viewModeSwitcher);
+
+        $viewModeSwitcher->on(
+            ViewModeSwitcher::ON_REQUEST,
+            function (
+                ServerRequestInterface $request,
+                ViewModeSwitcher $viewModeSwitcher
+            ) {
+                $viewModeSwitcher->populate([
+                    $viewModeSwitcher->getViewModeParam()
+                        => $request->getQueryParams()[$viewModeSwitcher->getViewModeParam()] ?? null
+                ]);
+
+                $this->emit(static::ON_VIEW_MODE_SET, [$viewModeSwitcher]);
+            }
+        );
+
+        $viewModeSwitcher->on(
+            ViewModeSwitcher::ON_SUBMIT,
+            function (
+                ViewModeSwitcher $viewModeSwitcher
+            ): void {
+                $previousViewModeName =
+                    $viewModeSwitcher->getRequest()->getQueryParams()[$viewModeSwitcher->getViewModeParam()]
+                    ?? $viewModeSwitcher->getDefaultViewMode()->getName();
+                $previousViewMode = $viewModeSwitcher->getViewModes()[$previousViewModeName];
+
+                if ($viewModeSwitcher->getViewMode()->getName() !== $previousViewMode->getName()) {
+                    $redirectUrl = Url::fromRequest();
+                    $this->emit(static::ON_VIEW_MODE_CHANGE, [$viewModeSwitcher, $previousViewMode, $redirectUrl]);
+                    $redirectUrl->setParam(
+                        $viewModeSwitcher->getViewModeParam(),
+                        $viewModeSwitcher->getViewMode()->getName()
+                    );
+                    $this->redirectNow($redirectUrl);
+                }
+            }
+        );
+
+        return $viewModeSwitcher;
+    }
+
+    /**
+     * Register the given control for lookup by {@see self::getTrackedControl()} and for {@see self::handleControls()}
+     *
+     * @param Form $control
+     *
+     * @return $this
+     */
+    protected function trackControl(Form $control): static
+    {
+        $this->trackedControls[] = $control;
+
+        return $this;
+    }
+
+    /**
+     * Get the tracked control of the given type
+     *
+     * @template TControl of Form
+     *
+     * @param class-string<TControl> $type
+     *
+     * @return ?TControl
+     */
+    protected function getTrackedControl(string $type): ?Form
+    {
+        foreach ($this->trackedControls as $control) {
+            if ($control instanceof $type) {
+                return $control;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * Call {@see Form::handleRequest()} on every control in {@see self::$trackedControls}.
+     *
+     * @param ServerRequestInterface $request
+     *
+     * @return $this
+     */
+    protected function handleControls(ServerRequestInterface $request): static
+    {
+        foreach ($this->trackedControls as $control) {
+            $control->handleRequest($request);
+        }
+
+        return $this;
+    }
+}
--- a/src/Compat/CompatController.php
+++ b/src/Compat/CompatController.php
@@ -12,10 +12,13 @@ use ipl\Html\HtmlString;
 use ipl\Html\ValidHtml;
 use ipl\Orm\Query;
 use ipl\Stdlib\Contract\Paginatable;
+use ipl\Stdlib\Events;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\PaginationControl;
 use ipl\Web\Control\SearchBar;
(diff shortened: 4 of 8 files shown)


src/Common/Controls.php
```diff
@@ -0,0 +1,75 @@
+<?php
```
sukhwinder33445, line 1: To create a view mode switcher, I used the following code you shared with me, in a controller:
```php
use Controls;
//...
$viewModeSwitcher->on(ViewModeSwitcher::ON_SUBMIT, function (ViewModeSwitcher $switcher): void {
    $this->redirectNow(
        Url::fromRequest()->setParam($switcher->getViewModeParam(), $switcher->getViewMode())
    );
});

$viewMode = $viewModeSwitcher->getViewMode();

$limitControl = new LimitControl(Url::fromRequest());
if ($viewMode === 'minimal') {
    $limitControl->setDefaultLimit(LimitControl::DEFAULT_LIMIT * 2);
}

$this->trackControl($limitControl);
$this->handleControls($this->getServerRequest());

//...
$this->addControl($limitControl);
$this->addControl($viewModeSwitcher);
```

This seems boilerplate code to write in each controller when we want a `ViewModeSwitcher`.
At least the adjustment of the default limit for `LimitControl` in minimal mode could be centralized. Also, the `PaginationControl` (page numbers) currently isn't adjusted when the user selects minimal mode, the limit switches from 25 to 50, but the number of pages stays the same.

sukhwinder33445 (commented):
The view mode switcher is no longer aligned to the right as it was before:
Before:
Image

Now:
Image

Please fix the alignment and style.

src/Common/Controls.php
BastianLedererIcinga: I adjusted the trait. Your consumer code should now look like this (pagination is optional):
```
$viewModeSwitcher = $this->createViewModeSwitcher($this->params);
$limitControl = $this->createLimitControl();
$paginationControl = $this->createPaginationControl($paginatable);

$this->applyViewModeLimit($viewModeSwitcher, $limitControl, $paginationControl);
$this->handleControls($this->getServerRequest());

$this->addControl($limitControl);
$this->addControl($viewModeSwitcher);
$this->addControl($paginationControl);
```

src/Common/Controls.php
```diff
@@ -0,0 +1,129 @@
...
+     * @param LimitControl $limitControl
+     * @param ?PaginationControl $paginationControl
+     *
+     * @return void
+     */
+    protected function applyViewModeLimit(
+        ViewModeSwitcher $viewModeSwitcher,
+        LimitControl $limitControl,
+        ?PaginationControl $paginationControl = null
+    ): void {
```
sukhwinder33445: Instead of passing the `viewModeSwitcher` as a parameter again, you can use the instance stored in the `$trackedControls` array.

src/Control/ViewModeSwitcher.php
```diff
@@ -0,0 +1,210 @@
...
+    public const DEFAULT_VIEW_MODE = 'common';
+
+    /** @var string Default view mode param */
+    public const DEFAULT_VIEW_MODE_PARAM = 'view';
+
+    /** @var array<string, string> View mode-icon pairs */
+    public static array $viewModes = [
+        'minimal' => 'minimal',
+        'common' => 'default',
+        'detailed' => 'detailed',
```
sukhwinder33445: Remove this comma please.

src/Common/Controls.php
sukhwinder33445: And please add test.

src/Common/Controls.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public function createViewModeSwitcher(
+        UrlParams $params,
+        ?string $viewModeParam = null,
+        string $viewModeSwitcherClass = ViewModeSwitcher::class
+    ): ViewModeSwitcher {
+        $viewModeSwitcher = new $viewModeSwitcherClass();
+        if ($viewModeParam !== null) {
+            $viewModeSwitcher->setViewModeParam($viewModeParam);
+        }
```
sukhwinder33445: The third parameter allows passing any class-string, which can lead to failure.
I'd suggest throwing an exception if the created object is not an instance of the `ViewModeSwitcher` class.

sukhwinder33445 (commented):
Apart from the changes I’ve pointed out, the code looks fine to me.

sukhwinder33445 (dismissed):

asset/css/controls.less
```diff
@@ -127,10 +127,12 @@
   }
 
   > .sort-control,
+  > .view-mode-switcher,
```
nilmerg, line 130: please update the figure above as well by including this widget

asset/css/view-mode-switcher.less
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+
+  input {
+    display: none;
+  }
+
+  label {
+    color: var(--control-color, @control-color);
+    line-height: 1;
+    background: var(--low-sat-blue, @low-sat-blue);
```
nilmerg: `@low-sat-blue` is Icinga Web specific. Use `@default-input-bg`

asset/css/view-mode-switcher.less
```diff
@@ -0,0 +1,42 @@
...
+      border-bottom-left-radius: 0.25em;
+    }
+
+    &:last-of-type {
+      border-top-right-radius: 0.25em;
+      border-bottom-right-radius: 0.25em;
+    }
+
+    &:not(:last-of-type) {
+      border-right: 1px solid var(--low-sat-blue-dark, @low-sat-blue-dark);
```
nilmerg: `@default-input-hover-bg`

asset/css/view-mode-switcher.less
```diff
@@ -0,0 +1,42 @@
...
+
+    i {
+      // fix height for Chrome
+      display: block;
+    }
+  }
+
+  input[checked] + label {
+    background-color: var(--control-color, @control-color);
+    color: var(--text-color-on-icinga-blue, @text-color-on-icinga-blue);
```
nilmerg: `@default-text-color-inverted`

src/Common/Controls.php
```diff
@@ -0,0 +1,157 @@
+<?php
+
+namespace ipl\Web\Common;
+
+use Icinga\Web\UrlParams;
```
nilmerg: This opposes the basic assumption to have an independent implementation. But this isn't specific to this change anyway, as nothing new outside the `ipl\Web\Compat` namespace must depend on it.

nilmerg (changes requested):
I'm not really convinced that this follows what I proposed in the issue. The view mode dependent limit change is not automatically done anymore, and the dynamic correction of the current page is missing. Custom rules, so to say, are also not even considered.

And where's:
> So the new implementation should allow to add new modes as well as to remove default modes.

It also misses basic accessibility support, as the id protection has no fallback. Since it is now a library part I consider this a must, as unique ID's are not a requirement of Icinga Web, only the part that it must be based on the container id. `ipl-html` also provides a fallback in case of the new decorator implementation for this very reason: A sane default.

So this is far from finished, I'm afraid.

src/Compat/CompatController.php
```diff
@@ -305,6 +310,42 @@ public function createSortControl(Query $query, array $columns): SortControl
...
+     * Does nothing on its own. Subclasses using the trait get its implementation instead.
+     * This only exists so that {@see self::createLimitControl()} and {@see self::createPaginationControl()}
+     * can call it unconditionally, which is what lets the {@see ViewModeSwitcher} adjust their page size.
+     *
+     * @param ValidHtml $control
+     * @param bool $handleRequest
+     *
+     * @return $this
+     */
+    protected function trackControl(ValidHtml $control, bool $handleRequest = false): static
```
BastianLedererIcinga: I didn't use the `class_uses()` guard you suggested in person, because it doesn't contain [traits used by a parent class.](https://www.php.net/manual/en/function.class-uses.php#:~:text=This%20does%20however%20not%20include%20any%20traits%20used%20by%20a%20parent%20class.)
And I think this is less ugly than a `method_exists()` guard

src/Common/Controls.php
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Provides factory methods to prepare reusable web controls and allows to handle their requests
+ *
+ * {@see static::handleControls()} can be called to call {@see Form::handleRequest()} for all Controls created by
+ * this trait.
+ *
+ * @phpstan-require-extends CompatController
+ */
+trait Controls
```
nilmerg, line 17: I think this should use the `Events` trait as it relies on it.

src/Common/Controls.php
```diff
@@ -0,0 +1,133 @@
...
+            ): void {
+                $previousViewMode = $viewModeSwitcher
+                    ->getRequest()
+                    ->getQueryParams()[$viewModeSwitcher->getViewModeParam()]
+                    ?? $viewModeSwitcher->getDefaultViewMode();
+                if ($viewModeSwitcher->getViewMode() !== $previousViewMode) {
+                    $redirectUrl = Url::fromRequest();
+                    $this->emit(static::ON_VIEW_MODE_CHANGE, [$viewModeSwitcher, $previousViewMode, $redirectUrl]);
+                    $redirectUrl->setParam($viewModeSwitcher->getViewModeParam(), $viewModeSwitcher->getViewMode());
+                    $this->redirectNow($redirectUrl);
```
nilmerg, line 89: this seems the only requirement for the controller extends, so this should be an abstract method here to force this combination.

src/Compat/CompatController.php
```diff
@@ -274,6 +300,44 @@ public function createPaginationControl(Paginatable $paginatable): PaginationCon
...
+            function (
+                ViewModeSwitcher $viewModeSwitcher,
+                string $previousViewMode,
+                Url $redirectUrl
+            ) use ($paginationControl) {
+                if (
+                    ! isset($this->getServerRequest()->getQueryParams()[$paginationControl->getPageSizeParam()])
+                    && $viewModeSwitcher->getViewMode() !== $previousViewMode
+                ) {
+                    $previous = $viewModeSwitcher->getPageSizeForViewMode($previousViewMode);
```
nilmerg: If that's the only reason why `getPageSizeForViewMode` exists, I'd rather suggest to provide a proper view mode instance instead of a string here (to the listener). Then it's always: `getViewMode()->getPageSize()`

src/Control/ViewModeSwitcher.php
```diff
@@ -0,0 +1,302 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Get the name of the active view mode
+     *
+     * @return string
+     */
+    public function getViewMode(): string
```
nilmerg: I think this should return the view mode instance. Not only because my other comment implies this already, it will probably also of use because the way how custom modes can be used it allows to pass extended classes and not only the base.

nilmerg (changes requested):

tests/Common/ControlsTest.php
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    protected function tearDown(): void
+    {
+        if ($this->queryString === null) {
+            unset($_SERVER['QUERY_STRING']);
+        } else {
+            $_SERVER['QUERY_STRING'] = $this->queryString;
+        }
+    }
```
nilmerg, line 70: As discussed, this is AI slop to me. There's no reason a unittest has to abide to something that isn't related to the actual test cases. It has to cleanup global state after it ran, but preserving global state isn't required usually.

nilmerg (changes requested):
Looks fine now, thank you. Though, I will first check any dependent implementation before merging it.

nilmerg (changes requested):

nilmerg (approved):
🎉

PR #396, opened 2026-07-13
CsrfCounterMeasure: Only reject unsafe requests upon validation

nilmerg:
Calling `$form->isValid()` explicitly right after construction throws an error as the csrf element is validated and has no value. Similarly, assembling a form with such an element should only fail if it is validated.

This does not fix an issue, it will only streamline the behavior. It also ensures expectations of poeple familiar with form request handling (like me) are met.

### DIFF
--- a/src/Common/CsrfCounterMeasure.php
+++ b/src/Common/CsrfCounterMeasure.php
@@ -49,23 +49,21 @@ trait CsrfCounterMeasure
      *
      * @return FormElement
      *
-     * @throws Error If {@see requestIsSafe()} returns false
-     *
      * @deprecated Use {@see addCsrfCounterMeasure()} instead
      */
     protected function createCsrfCounterMeasure($uniqueId)
     {
         $requestIsSafe = $this->requestIsSafe();
-
         if ($requestIsSafe !== null) {
-            if (! $requestIsSafe) {
-                throw new Error('Rejecting cross-site request');
-            }
-
             return new HiddenElement('CSRFToken', [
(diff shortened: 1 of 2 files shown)


Al2Klimov (dismissed):

src/Common/CsrfCounterMeasure.php
```diff
@@ -49,19 +49,13 @@ public function disableCsrfCounterMeasure(): void
...
      * @deprecated Use {@see addCsrfCounterMeasure()} instead
      */
     protected function createCsrfCounterMeasure($uniqueId)
     {
         $requestIsSafe = $this->requestIsSafe();
 
-        if ($requestIsSafe !== null) {
-            if (! $requestIsSafe) {
-                throw new Error('Rejecting cross-site request');
-            }
```
Al2Klimov, line 63: If you're not in a hurry, you might also put this exception in a dedicated validator which works just like the `return true;` below, but instead throws the exception. This way you don't have to compute seed and stuff.

TheSyscall (dismissed):

src/Common/CsrfCounterMeasure.php
nilmerg: If I'd do this, the next request (actual submission) has no token and fails.

nilmerg: But if the request is safe/unsafe, it's not required…

Al2Klimov (approved):

PR #397, opened 2026-07-17
Callout: Normalize text content spacing

TheSyscall:
Move `.callout-title` inside `.callout-text` to scope it correctly, and add rules to strip default margin from direct children of `.callout-text`.

This improves the display of lists and paragraphs.

**before**
Screenshot From 2026-07-17 09-18-48

**after**
Screenshot From 2026-07-17 09-16-00

### DIFF
--- a/asset/css/callout.less
+++ b/asset/css/callout.less
@@ -16,13 +16,26 @@
     margin: 0;
   }
 
-  .callout-title {
-    margin-bottom: .5em;
-  }
-
   .callout-text {
     display: flex;
     flex-direction: column;
+
+    & > * {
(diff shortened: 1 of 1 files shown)


flourish86 (dismissed):
"Yes" 👍

jrauh01 (commented):
I tried it and it works fine. However, we allow `ValidHtml` to be passed as content, which may be a manually styled html element. The new selector would strip the margin from any direct child, even elements that were given a margin intentionally.

jrauh01 (approved):
Explicitly added margins of elements inside the callout are now preserved.

PR #398, opened 2026-07-17
Override `onError()` in `CompatForm` for prettier form errors

jrauh01:
Each form error is now shown as a separate `Callout` instead of collecting all messages into a single `<ul class="errors">`. This displays the errors in a prettier way for the user.

Before
Screenshot 2026-07-17 at 15 11 27

After
Screenshot 2026-07-17 at 15 11 40

### DIFF
--- a/asset/css/compat.less
+++ b/asset/css/compat.less
@@ -148,6 +148,11 @@ form.icinga-form .suggestion-element-group {
   }
 }
 
+// Form error callout styles
+form > .error-callouts > .callout-type-error:not(:last-child) {
+  margin-bottom: .5em;
+}
+
 // Display form element styles
 form.icinga-form .control-group > .display-form-element {
   flex: 1 1 auto;
--- a/src/Compat/CompatForm.php
+++ b/src/Compat/CompatForm.php
@@ -6,6 +6,7 @@ use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
 use Icinga\Exception\IcingaException;
 use InvalidArgumentException;
+use ipl\Html\Attributes;
 use ipl\Html\Contract\FormElement;
 use ipl\Html\Contract\FormSubmitElement;
 use ipl\Html\Contract\HtmlElementInterface;
@@ -13,11 +14,14 @@ use ipl\Html\Form;
 use ipl\Html\FormElement\SubmitButtonElement;
 use ipl\Html\FormElement\SubmitElement;
 use ipl\Html\HtmlDocument;
+use ipl\Html\HtmlElement;
 use ipl\Html\HtmlString;
 use ipl\I18n\Translation;
+use ipl\Web\Common\CalloutType;
 use ipl\Web\Compat\FormDecorator\PrimaryButtonDecorator;
 use ipl\Web\FormDecorator\IcingaFormDecorator;
 use ipl\Web\Compat\FormDecorator\LabelDecorator;
+use ipl\Web\Widget\Callout;
 use Stringable;
 use Throwable;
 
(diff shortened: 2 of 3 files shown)


flourish86 (dismissed):
Yes!

src/Compat/CompatForm.php
```diff
@@ -223,4 +227,30 @@ protected function logAndShowError(Throwable|string $error, string $template, mi
...
+    /**
+     * Render form error messages as callouts
+     *
+     * Overrides the default rendering, which produces a single `<ul class="errors">`
+     * list, to instead prepend a `<span class="error-callouts"></span>` containing
+     * one error callout per message.
+     *
+     * @return void
+     */
+    protected function onError(): void
```
lippserd: Adding a return type to this protected extension hook causes a fatal error for
existing overrides that omit it, including `Jira\Web\Form\NewIssueForm`. Keep
the inherited hook untyped.

```suggestion
    protected function onError()
```

src/Compat/CompatForm.php
```diff
@@ -223,4 +227,30 @@ protected function logAndShowError(Throwable|string $error, string $template, mi
...
+    {
+        $errors = new HtmlElement('span', new Attributes(['class' => 'error-callouts']));
+
+        foreach ($this->getMessages() as $message) {
+            if ($message instanceof Throwable) {
+                $message = $message->getMessage();
+            }
+
+            $errors->addHtml(new Callout(CalloutType::Error, $message));
```
lippserd: This `<span>` contains `<div>` callouts, which is invalid HTML, and the previous
list semantics are lost. Use a flow-content wrapper and expose the messages as
an ARIA list. Update the PHPDoc's `<span>` wording to match.

```suggestion
        $errors = new HtmlElement('div', new Attributes([
            'class' => 'error-callouts',
            'role' => 'list',
        ]));

        foreach ($this->getMessages() as $message) {
            if ($message instanceof Throwable) {
                $message = $message->getMessage();
            }

            $errors->addHtml(
                (new Callout(CalloutType::Error, $message))
                    ->setAttribute('role', 'listitem')
            );
```

asset/css/compat.less
```diff
@@ -148,6 +148,11 @@ form.icinga-form .suggestion-element-group {
   }
 }
 
+// Form error callout styles
+form.icinga-form > .error-callouts > .callout-type-error {
```
lippserd: `CompatForm` subclasses can replace the default `icinga-form` class, so this
selector leaves their callouts unspaced. Target the generated wrapper
independently of that optional class.

```suggestion
form > .error-callouts > .callout-type-error {
```

lippserd (changes requested):
Requesting changes for the compatibility, markup semantics, and selector-scope
issues called out inline.

Please add regression coverage for:

- Multiple form-level messages render in one wrapper and remain in source order.
  The initial implementation reversed them, so this protects the correction made
  later in the PR.
- HTML-like message text remains escaped. The new callout rendering path must not
  turn validation content into executable markup.
- No wrapper is emitted when there are no messages. This protects the empty-state
  markup from unnecessary containers.

src/Compat/CompatForm.php
```diff
@@ -223,4 +227,36 @@ protected function logAndShowError(Throwable|string $error, string $template, mi
         $this->addMessage(str_replace('{error}', $errorMessage, $template), ...$args);
         $this->onError();
     }
+
+    /**
+     * Render form error messages as callouts
+     *
+     * Overrides the default rendering, which produces a single `<ul class="errors">`
+     * list, to instead prepend a `<span class="error-callouts"></span>` containing
```
lippserd: The previous inline request asked to update the PHPDoc alongside the element change. The `<span>` reference here was not updated:


```suggestion
     * list, to instead prepend a `<div class="error-callouts"></div>` containing
```

asset/css/compat.less
```diff
@@ -148,6 +148,11 @@ form.icinga-form .suggestion-element-group {
   }
 }
 
+// Form error callout styles
+form > .error-callouts > .callout-type-error {
```
lippserd: If the trailing margin below the last callout is not intentional, add `:not(:last-child)`:


```suggestion
form > .error-callouts > .callout-type-error:not(:last-child) {
```

tests/Compat/CompatFormTest.php
```diff
@@ -322,6 +322,78 @@ public function testFormControlsDecoration(): void
         $this->assertHtml($expected, $form);
     }
 
+    public function testOnErrorRendersMultipleMessagesInSourceOrder(): void
```
lippserd, line 325: This test runs against an empty form, so `prependHtml()` and `addHtml()` produce identical output.  A regression swapping them in the implementation would still pass this test. Add an element to the form before calling `triggerOnError()` and assert that `<div class="error-callouts">`appears before it in the rendered output.

lippserd (changes requested):

tests/Compat/CompatFormTest.php
jrauh01: Added a dedicated test to ensure the message are rendered above the elements.

lippserd (approved):

PR #403, opened 2026-07-31
Defer CSRF counter measure ID validation

lippserd:
Require an ID only when request metadata cannot establish its origin and token-based validation is necessary. Safe and cross-site requests can then take their existing paths without a configured ID.

### DIFF
--- a/src/Common/CsrfCounterMeasure.php
+++ b/src/Common/CsrfCounterMeasure.php
@@ -43,9 +43,9 @@ trait CsrfCounterMeasure
      *
      * If the {@see requestIsSafe()} check concludes the request is safe, returns a dummy element that accepts any
      * value. If it concludes the request is unsafe, throws an {@see Error}. If the check is inconclusive (legacy
-     * browser), creates a token-based CSRF element validated against $uniqueId.
+     * browser), requires $uniqueId and creates a token-based CSRF element validated against it.
      *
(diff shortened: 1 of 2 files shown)


src/Common/CsrfCounterMeasure.php
```diff
@@ -68,6 +68,10 @@ protected function createCsrfCounterMeasure($uniqueId)
             ]);
         }
 
+        if ($uniqueId === null) {
+            throw new Error('No CSRF counter measure ID set');
```
jrauh01, line 72: Please add `@throws` to document this error.

tests/Common/CsrfCounterMeasureTest.php
```diff
@@ -187,6 +226,24 @@ protected function assemble(): void
         };
     }
 
+    private function makeFormWithoutId(bool $disabled = false): Form
```
jrauh01, line 229: Personally I think you should remove the `$disabled` param and explicitly disable CSRF protection in the test itself via:
```php
$form = $this->makeFormWithoutId();
$form->disableCsrfCounterMeasure();
```
It's only used once anyway, and this would make it clearer at the call site.

jrauh01 (changes requested):
While I think it's an improvement on one hand to only check for an ID when it is really used, there is still the caveat that this could lead to a working form on the developer's browser, failing at runtime on the user's system.

PR #405, opened 2026-08-05
Fix `CsrfCounterMeasure` bypass for crafted cross-site requests

jrauh01:
The `Sec-Fetch-Site` based element decided the cross-site rejection from the submitted `CSRFToken` value instead of the server-computed result, so a crafted request could bypass the countermeasure in two ways:

1. An empty value nulled the element, and since it was not required its validator was skipped entirely and the request treated as valid.
2. Any other non-empty value was coerced to the validator's bool parameter, where every truthy string passed as safe.

The element now ignores the submitted value: it always reports a value so the validator runs regardless of input, and the validator rejects based solely on the `Sec-Fetch-Site` result. This restores the invariant that a cross-site request cannot be accepted no matter what the client sends.

### DIFF
--- a/tests/Common/CsrfCounterMeasureTest.php
+++ b/tests/Common/CsrfCounterMeasureTest.php
@@ -97,6 +97,19 @@ class CsrfCounterMeasureTest extends TestCase
         $form->isValid();
     }
 
+    public function testCrossSiteRequestIsRejectedWithEmptyToken(): void
+    {
+        $_SERVER['HTTP_SEC_FETCH_SITE'] = 'cross-site';
+
+        $this->expectException(Error::class);
+        $this->expectExceptionMessage('Rejecting cross-site request');
+
+        $form = $this->makeForm();
+        $form->populate(['CSRFToken' => '']);
+        $form->ensureAssembled();
+        $form->isValid();
+    }
+
     public function testCreateReturnsDummyElementForSafeRequest(): void
     {
         $_SERVER['HTTP_SEC_FETCH_SITE'] = 'same-origin';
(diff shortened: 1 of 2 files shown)


tests/Common/CsrfCounterMeasureTest.php
```diff
@@ -84,6 +84,32 @@ public function testUnsafeCrossSiteRequestIsRejected(): void
         $this->makeForm()->handleRequest($this->requestMock('POST'));
     }
 
+    public function testCrossSiteRequestIsRejectedWithNonEmptyToken()
```
lippserd: Missing `: void` return type. All recent test methods in this file declare it.

```suggestion
    public function testCrossSiteRequestIsRejectedWithNonEmptyToken(): void
```

tests/Common/CsrfCounterMeasureTest.php
```diff
@@ -84,6 +84,32 @@ public function testUnsafeCrossSiteRequestIsRejected(): void
...
+        $this->expectException(Error::class);
+        $this->expectExceptionMessage('Rejecting cross-site request');
+
+        $form = $this->makeForm();
+        $form->populate(['CSRFToken' => 'anything']);
+        $form->ensureAssembled();
+        $form->isValid();
+    }
+
+    public function testCrossSiteRequestIsRejectedWithEmptyToken()
```
lippserd: Missing `: void` return type. All recent test methods in this file declare it.

```suggestion
    public function testCrossSiteRequestIsRejectedWithEmptyToken(): void
```

lippserd (changes requested):
On top of the changes requested, please solve the following:

The PR description ends mid-sentence, at "This restores the invariant that a
cross-site". Please complete it.

In the same paragraph, the character before `Sec-Fetch-Site` is an acute accent
(´) rather than a backtick, which breaks the inline code span.

lippserd (approved):

